PR #129, opened 2024-10-01
Forbid users to claim tributes that belong to other proposals than the one user voted on

dusan-maksimovic:


contracts/tribute/src/testing.rs
```diff
@@ -1179,3 +1202,115 @@ fn test_query_outstanding_tribute_claims() {
         }
     }
 }
+
```
p-offtermatt, line 1205: do we even still need this then if we have the other test case you added? do you think it adds value?

p-offtermatt: ok, let's remove it!

dusan-maksimovic: Done in https://github.com/informalsystems/hydro/pull/129/commits/9a252f41b8a55c04be3951b9ea7cb3ea1f0ca072

p-offtermatt (approved):

PR #134, opened 2024-10-07
Adapt tests after code changes

stana-miric:
This pr adapts test after the code is changed to:
- increase power of the proposal right after the locking(no need of voting action)
- not use community tax


dusan-maksimovic (approved):
LGTM! I think we should trigger this tests once per day, since they take bit more time to run. I will take care of this in a separate PR.

PR #135, opened 2024-10-08
fix: do not insert proposal into PROPS_BY_SCORE on proposal creation

dusan-maksimovic:
This PR removes code that inserts the proposal into PROPS_BY_SCORE during create_proposal() execution. Since the proposal initially has voting power 0, it should not be inserted into this storage. This is also aligned with what we do in vote() when user switches the vote: if the old proposal's power drops to 0, it will be removed from the PROPS_BY_SCORE map.


p-offtermatt (approved):

p-offtermatt:
hey, add a changelog entry please :)

PR #137, opened 2024-10-09
Add pilot round lock duration

p-offtermatt:


dusan-maksimovic (approved):
LGTM!

Changelog should be added.

contracts/hydro/src/testing.rs
```diff
@@ -1837,3 +1838,86 @@ fn assert_proposal_voting_power(
...
+        );
+
+        let lock_msg = ExecuteMsg::LockTokens {
+            lock_duration: case.lock_duration,
+        };
+
+        let res = execute(deps.as_mut(), env.clone(), info.clone(), lock_msg.clone());
+
+        if case.expect_error {
+            assert!(
```
dusan-maksimovic, line 1909: Perhaps also check that the error message contains "Lock duration must be 1 epoch" string?

p-offtermatt: Thanks, done!

PR #146, opened 2024-10-15
Adjust tribute refunding condition

### LINKED ISSUE #140: Adjust refunding condition
We want to add an extra condition on tributes.
In particular, we would like to not pay out tributes but instead refund them when they obtain *less than some percentage of the total votes*.
In some contexts, this is a more natural condition for refunding tributes (because positions that would gain very little liquidity are not worth deploying, as they are a lot of effort for the Hydro DAO to manage, we don't want those positions to get deployed, thus we would like to refund those tributes).

dusan-maksimovic:
Closes #140

This PR does the following:
- Tributes can be claimed only if the proposal was among the top N and it received at least the specified percentage threshold of the total voting power.
- Tributes can be refunded only if the proposal wasn't among the top N or it was among the top N but it received less than the specified percentage threshold of the total voting power.
- Percentage threshold is specified through the instantiation message. For the deployed contracts, a proper migration is added that allows this parameter to be specifed.


p-offtermatt (commented):
I still think we should extract a method from refund_tribute and claim_tribute.
These two going out of sync is the kind of thing that will cause bugs in the future.

How about making it a function "TributeIsRefundable() -> (bool, str)", where the string is a message that explains the "why" and that can be used in error returns.
For example, if the proposal is in the top N and has the percentage, the return is:
false, "proposal is in the top N and has at least x% of the votes".
if its not in the top N:
true, "proposal is not in the top N"

if it does not have the percentage threshold:
true, "proposal does not have at least x% of the votes"


Then the message can be used to inform the error return. so for claiming, if "IsRefundable" returns true, it would go:
"Cannot claim tribute:" + message, so e.g. it would look like
"Cannot claim tribute: proposal is not in the top N"

contracts/tribute/src/contract.rs
```diff
@@ -335,6 +326,62 @@ fn refund_tribute(
...
+impl TopNProposalInfo {
+    fn are_tributes_claimable(&self) -> Result<Proposal, ContractError> {
+        match self.top_n_proposal.as_ref() {
+            None => Err(ContractError::Std(StdError::generic_err(
+                "User voted for proposal outside of top N proposals",
+            ))),
+            Some(proposal) => {
+                if !self.is_above_voting_threshold {
+                    return Err(ContractError::Std(StdError::generic_err(
+                        "Tribute can not be claimed because the proposal received less voting percentage than required to deploy the liquidity.")));
```
p-offtermatt, line 343: I would abstract away a bit the reasoning behind it and just say "Tribute not claimable: Proposal received less voting percentage than threshold: X required, but is Y"

for our code, the fact that this is about deploying liquidity shouldn't matter, our contract just has proposals that people vote on and will get tribute, or won't get tribute

contracts/tribute/src/contract.rs
```diff
@@ -335,6 +326,62 @@ fn refund_tribute(
         }))
 }
 
+struct TopNProposalInfo {
+    pub top_n_proposal: Option<Proposal>,
+    pub is_above_voting_threshold: bool,
+}
```
p-offtermatt, line 332: Would recommend adding quick docstrings to the struct and the fields,
explaining what to expect (in particular saying that it will be None, false if the prop is not in the top N)

contracts/tribute/src/msg.rs
```diff
@@ -27,4 +29,6 @@ pub enum ExecuteMsg {
 }
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
-pub struct MigrateMsg {}
+pub struct MigrateMsg {
+    pub min_prop_percent_to_deploy: Uint128,
```
p-offtermatt, line 33: Maybe we can also leave out the deploying part here and just make it about refunding vs claiming tributes. something like "min_prop_percent_for_claimable_tributes"

contracts/tribute/src/msg.rs
```diff
@@ -1,10 +1,12 @@
+use cosmwasm_std::Uint128;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
 pub struct InstantiateMsg {
     pub hydro_contract: String,
     pub top_n_props_count: u64,
+    pub min_prop_percent_to_deploy: Uint128,
```
p-offtermatt, line 9: see below on naming

contracts/tribute/src/contract.rs
```diff
@@ -647,7 +694,34 @@ fn get_top_n_proposals(
 }
 
 #[cfg_attr(not(feature = "library"), entry_point)]
-pub fn migrate(_deps: DepsMut, _env: Env, _msg: MigrateMsg) -> Result<Response, ContractError> {
+pub fn migrate(deps: DepsMut, _env: Env, msg: MigrateMsg) -> Result<Response, ContractError> {
```
p-offtermatt, line 697: wdyt about putting this into a separate "migrate.rs" file? I want to start splitting up the main contract files more, since they tend to get pretty big with our current conventions

contracts/tribute/src/contract.rs
```diff
@@ -647,7 +694,34 @@ fn get_top_n_proposals(
...
+    let contract_version = get_contract_version(deps.storage)?;
+    if contract_version.version == CONTRACT_VERSION {
+        return Err(ContractError::Std(StdError::generic_err(
+            "Contract is already migrated to the newest version.",
+        )));
+    }
+
+    if msg.min_prop_percent_to_deploy > Uint128::new(100) {
+        return Err(ContractError::Std(StdError::generic_err(
+            "Minimum proposal percentage to deploy the liquidity must be between 0 and 100.",
```
p-offtermatt, line 708: see other comments re naming/"deploying"

p-offtermatt (approved):

PR #147, opened 2024-10-15
Allow tributes for proposals in past rounds

p-offtermatt:
When adding a tribute, the round id is now an input parameter, instead of computed from the current time.
This means that tribute can even be added for tributes in past rounds (before, the round id being automatically computed would cause the prop not to be queried from the store)


dusan-maksimovic (approved):
LGTM!

Do you think it would be useful to add info to each Tribute that would say in which round the tribute was added? The motivation would be: if someone wants to mess around with tributes for some proposal, this way we could at least know if some tribute was added before or after the voting was finished. WDYT?

p-offtermatt:
Hm, it's not a bad idea. It might be even more helpful to keep an exact "creation time" timestamp though; that could help with showing graphs of how the tribute evolved over time, if we ever need that

PR #156, opened 2024-10-21
Add a migration to push the start date to be later

p-offtermatt:
## Description

Closes: N/A

Adds a migration to push the start date to be later.
Concretely:
the migrate message will contain the new first round start date.
The contract will simply set its start date to that date, and set the end of all existing locks to the end of the new first round.

The migration has a bunch of checks to make sure its assumptions are correct:
* it only works while the contract is in its first round
* it only works if the new first round end is in the future

---

### Author Checklist

I have...

* [o] Added/adjusted the necessary [interchain tests](./test/interchain/)
* [o] Added a changelog entry in `.changelog`


contracts/hydro/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "hydro"
-version = "1.0.0"
+version = "1.1.0"
```
dusan-maksimovic, line 3: Should we also use version 1.1.0 for tribute contract? Currently it is set to 2.0.

dusan-maksimovic (approved):
Missing changelog, otherwise good.

contracts/hydro/Cargo.toml
p-offtermatt: yes, probably right

Issue #125, opened 2024-09-24
Allowing fractional votes


p-offtermatt:
A user should be able to split their voting power between multiple different proposals.

This is important for users that have significant voting power:
A single user that has a lot of voting power (e.g. a vote aggregator) might not want to throw all their weight behind one proposal.

PR #157, opened 2024-10-21
Fractional voting support

dusan-maksimovic:
## Description

Closes: #125 

Implements fractional voting by allowing users to specify lock ids they want to use to vote for one or more proposals in the round and tranche.

---

### Author Checklist

I have...

### DIFF
--- a/contracts/hydro/src/testing.rs
+++ b/contracts/hydro/src/testing.rs
@@ -2173,7 +2173,7 @@ fn test_get_vote_for_update() {
     let prop_id_2 = 1;
 
     let validator_1 = String::from(VALIDATOR_1);
-    let validator_2 = String::from(VALIDATOR_1);
+    let validator_2 = String::from(VALIDATOR_2);
 
     let lockup_id_1 = 0;
     let lockup_id_2 = 1;


contracts/hydro/src/contract.rs
```diff
@@ -676,135 +675,210 @@ fn vote(
...
+                        lock_ids: unique_lock_ids,
+                    })
+                } else {
+                    None
+                }
+            } else {
+                None
+            }
+        })
+        .collect();
```
p-offtermatt, line 706: what happens when a lock_id is mentioned twice?

dusan-maksimovic: It will be taken into account only the first time it is mentioned (see line 691). Personally, I think we should completely reject such transactions, but that isn't the approach we have taken so far, and therefore I implemented the filtering.

p-offtermatt: I think we should reject the transaction, because there is no obvious way to determine which was the intended vote if it appears twice.

p-offtermatt: If it doesn't make the code less elegant, let's go for that approach

contracts/hydro/src/contract.rs
```diff
@@ -676,135 +675,210 @@ fn vote(
...
+    for proposal_to_lockups in proposals_votes {
+        let proposal_id = proposal_to_lockups.proposal_id;
 
-    for lock in locks {
-        let (_, lock_entry) = lock?;
+        // TODO: optimize so that proposal related stores are updated only once
+        // TODO: time_weighted_shares on Vote will have only one item
+        for lock_id in proposal_to_lockups.lock_ids {
+            // Get sender's locked shares for each validator
+            let mut time_weighted_shares_map: HashMap<String, Decimal> = HashMap::new();
```
p-offtermatt, line 792: this map seems unnecessary now, because each lock_id will have only shares from one validator in it

contracts/hydro/src/contract.rs
```diff
@@ -676,135 +675,210 @@ fn vote(
...
-            round_id,
-            proposal_id,
-            validator.to_string(),
-            *num_shares,
-        )?;
-    }
+            // if the user doesn't have any shares that give voting power, we don't need to do anything
+            if time_weighted_shares_map.is_empty() {
+                return Ok(response);
+            }
```
p-offtermatt, line 835: remove, since the map will have at most one entry anyways

p-offtermatt, line 835: Remove. the map will have exactly one entry when we get here, anyways

contracts/hydro/src/contract.rs
```diff
@@ -676,135 +675,210 @@ fn vote(
...
+            for (validator, num_shares) in time_weighted_shares_map.iter() {
+                // add the validator shares to the proposal
+                add_validator_shares_to_proposal(
+                    deps.storage,
+                    round_id,
+                    proposal_id,
+                    validator.to_string(),
+                    *num_shares,
+                )?;
+            }
```
p-offtermatt, line 847: same, there will only be 1 entry

contracts/hydro/src/contract.rs
dusan-maksimovic: Done in https://github.com/informalsystems/hydro/pull/157/commits/ece382d60f90a320440f6d84ef8270d3cc632f64

contracts/hydro/src/contract.rs
dusan-maksimovic: Done in https://github.com/informalsystems/hydro/pull/157/commits/21e74b10f1675c18a3fa0df0beb1d1c15f0458b7

contracts/hydro/src/contract.rs
dusan-maksimovic: Done in https://github.com/informalsystems/hydro/pull/157/commits/21e74b10f1675c18a3fa0df0beb1d1c15f0458b7

contracts/hydro/src/contract.rs
dusan-maksimovic: Done in https://github.com/informalsystems/hydro/pull/157/commits/21e74b10f1675c18a3fa0df0beb1d1c15f0458b7

contracts/hydro/src/contract.rs
```diff
@@ -676,135 +675,186 @@ fn vote(
...
+
+            // Delete vote
+            VOTE_MAP.remove(
+                deps.storage,
+                ((round_id, tranche_id), info.sender.clone(), lock_id),
+            );
+
+            response = response.add_attribute("old_proposal_id", vote.prop_id.to_string());
```
p-offtermatt, line 775: this should be refined to include the lock_id as part of the response, right?

dusan-maksimovic: Added in https://github.com/informalsystems/hydro/pull/157/commits/f3f442ab6582ac8acd2ea812322edc73a28e7d10

p-offtermatt (approved):

contracts/hydro/src/contract.rs
```diff
@@ -1413,25 +1466,52 @@ pub fn query_user_voting_power(
     Ok(UserVotingPowerResponse { voting_power })
 }
 
-pub fn query_user_vote(
+// This function queries user votes for the given round and tranche.
+// It goes through all user votes per lock_id and groups them by the
+// proposal ID. The returned result will contain one VoteWithPower per
+// each proposal ID, with the total power sumed up from all lock IDs
```
p-offtermatt, line 1472:
```suggestion
// each proposal ID, with the total power summed up from all lock IDs
```

contracts/hydro/src/contract.rs
```diff
@@ -1770,6 +1863,70 @@ fn update_voting_power_on_proposals(
...
+// more tokens or refreshes the existing lockup. Whether some vote should be
+// updated or not is determined by the following logic:
+//
+// If user is refreshing a lock, we check if this lock was already voted with.
+// If yes, this vote will be updated. If no, we will not add a vote for it. Instead,
+// we leave user a choice to vote later.
+// Is user is adding a new lock, we check if the user already voted for some proposals.
+// If user voted for only one proposal, then the new lock power will be added to it.
+// If user voted for multiple proposals, then we do not add a vote for new lock entry.
+fn get_vote_for_update(
```
p-offtermatt, line 1876: this could probably use its own unit test; the logic above is already a pretty good guide as to what to test

contracts/tribute/src/contract.rs
```diff
@@ -173,31 +173,34 @@ fn claim_tribute(
...
         voter.clone().to_string(),
-    )?;
+    )?
+    .into_iter()
+    .find(|vote| vote.prop_id == tribute.proposal_id)
+    {
+        None => {
+            // Error out if user didn't vote for the proposal that the given tribute belongs to.
+            return Err(ContractError::Std(StdError::generic_err(
+                "User voted for a different proposal than the one this tribute belongs to",
```
p-offtermatt, line 192:
```suggestion
                "User didn't vote for the proposal this tribute belongs to",
```
since a user can vote for multiple proposals now, "a different proposal" doesn't seem quite right

contracts/hydro/src/contract.rs
dusan-maksimovic: Done in https://github.com/informalsystems/hydro/pull/157/commits/5158433fc44926ddc3da7c85ce47c05b7c15ccbe

dusan-maksimovic:
> Some comments. Could you add an extra test case to the tribute contract where a user votes for multiple different proposals and claims tributes for them? afaict that case is untested right now

Done in https://github.com/informalsystems/hydro/pull/157/commits/5158433fc44926ddc3da7c85ce47c05b7c15ccbe

contracts/hydro/src/testing.rs
```diff
@@ -2159,3 +2163,132 @@ fn test_refresh_multiple_locks() {
...
+#[test]
+fn test_get_vote_for_update() {
+    let round_id = 0;
+    let tranche_id = 0;
+
+    let prop_id_1 = 0;
+    let prop_id_2 = 1;
+
+    let validator_1 = String::from(VALIDATOR_1);
+    let validator_2 = String::from(VALIDATOR_1);
```
p-offtermatt, line 2176:
```suggestion
    let validator_2 = String::from(VALIDATOR_2);
```

p-offtermatt (approved):
Nice refactor in testing.rs, looks much nicer.
Just one last comment about a typo, but apart from that, LGTM to merge

Issue #154, opened 2024-10-16
Minimum liquidity request for tributes


p-offtermatt:
Each tribute should be able to have a "minimum liquidity request".
If the proposal doesn't get at least that amount of tribute, the tribute should be refundable.

Technically, maybe we could have the committee submit the amount of liquidity that will be deployed in each tranche in total to the contract before the end of each round; the tribute contract can simply use that and the vote percentage of each proposal to figure out how much liquidity would be exported to each prop, and make tribute refundable as appropriate

PR #162, opened 2024-10-29
Add a minimum export floor

p-offtermatt:
## Description

Closes: #154 

This PR does the following:
* modify the `Proposal` struct to include a `minimum_atom_liquidity_request` field. This field isn't really used anywhere within the code, but it seems useful to have this separated out (instead of e.g. in the description text), since it is something that should be set when the proposal is created and not changed anymore afterwards (so e.g. having it in the Github is not a good solution)
*  Add a map of `LiquidityDeployments` in the store: These are indexed by (round, tranche, proposal) and describe what liquidity is deployed to given proposals. The map can be managed by whitelist admins.
* Tributes have a new condition:
  * Tributes are refunable *only if* there exists a deployment for their prop in the store, and that proposal received zero liquidity
  * Tributes are claimable *only if* there exists a deployment for their prop in the store, and that proposal received non-zero liquidity

The new workflow is this:
After a round, the committee will, for each proposal, issue a transaction that will add a liquidity deployment to the proposal, proportionate to how much liquidity will be deployed to the proposal (or 0 if no liquidity is allocated)
Once that has been entered, tributes can be refunded/claimed as normal.

This PR thus solves the minimum export floor issue: projects whose minimum request is not satisfied will not get liquidity, so the whitelist admins will enter a corresponding deployment, and thus the tribute will be refundable

There are a few things to consider now:
* Can we connect the liquidity deployments with the minimum request to automatically decide whether tributes are refundable? => Automatically calculating minimum deployments to compare to the minimum requestb doesn't make sense - the committee should never deploy less liquidity than the minimum request. It is also really complicated to compare these in general (because the minimum request is in ATOM-equivalent, but deployed funds might be things like dATOM or stATOM, so we'd need a price oracle). just checking whether bids got liquidity at all is sufficient in practice
* Do we even need the top N and minimum voting percentage solutions for whether tributes are claimable anymore? => I think these are just superseded by the committee deciding to deploy or not deploy liquidity to certain things. I think we can replace the conditions in the tributes with queries that allow you to see which props clear the voting percentage threshold or are in the top N

---

### Author Checklist

I have...

### DIFF
--- a/contracts/hydro/schema/hydro_full_schema.json
+++ b/contracts/hydro/schema/hydro_full_schema.json
@@ -1152,6 +1152,101 @@
   "migrate": {},
   "sudo": {},
   "responses": {
+    "liquidity_deployment": {
+      "$schema": "http://json-schema.org/draft-07/schema#",
+      "title": "LiquidityDeploymentResponse",
+      "type": "object",
+      "required": [
+        "liquidity_deployment"
+      ],
+      "properties": {
+        "liquidity_deployment": {
+          "$ref": "#/definitions/LiquidityDeployment"
+        }
+      },
+      "additionalProperties": false,
+      "definitions": {
+        "Coin": {
+          "type": "object",
+          "required": [
+            "amount",
+            "denom"
+          ],
+          "properties": {
+            "amount": {
+              "$ref": "#/definitions/Uint128"
+            },
+            "denom": {
+              "type": "string"
+            }
+          },
+          "additionalProperties": false
+        },
+        "LiquidityDeployment": {
+          "type": "object",
+          "required": [
+            "deployed_funds",
+            "destinations",
+            "funds_before_deployment",
+            "proposal_id",
+            "remaining_rounds",
+            "round_id",
+            "total_rounds",
+            "tranche_id"
+          ],
+          "properties": {
+            "deployed_funds": {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/Coin"
+              }
+            },
+            "destinations": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            "funds_before_deployment": {
+              "type": "array",
+              "items": {
+                "$ref": "#/definitions/Coin"
+              }
+            },
+            "proposal_id": {
+              "type": "integer",
+              "format": "uint64",
+              "minimum": 0.0
+            },
+            "remaining_rounds": {
+              "type": "integer",
+              "format": "uint64",
+              "minimum": 0.0
+            },
+            "round_id": {
+              "type": "integer",
+              "format": "uint64",
+              "minimum": 0.0
+            },
+            "total_rounds": {
+              "type": "integer",
+              "format": "uint64",
+              "minimum": 0.0
+            },
+            "tranche_id": {
+              "type": "integer",
+              "format": "uint64",
+              "minimum": 0.0
+            }
+          },
+          "additionalProperties": false
+        },
+        "Uint128": {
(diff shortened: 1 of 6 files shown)


contracts/hydro/src/msg.rs
```diff
@@ -89,6 +90,23 @@ pub enum ExecuteMsg {
...
+
+    AddLiquidityDeployment {
+        round_id: u64,
+        tranche_id: u64,
+        proposal_id: u64,
+        destinations: Vec<String>,
+        deployed_funds: Vec<Coin>,
+        funds_before_deployment: Vec<Coin>,
+        total_rounds: u64,
+        remaining_rounds: u64,
```
dusan-maksimovic, line 102: The meaning of this fields should be documented somewhere.

contracts/hydro/src/query.rs
```diff
@@ -189,3 +207,13 @@ pub struct ValidatorPowerRatioResponse {
...
     pub managers: Vec<Addr>,
 }
+
+#[cw_serde]
+pub struct LiquidityDeploymentResponse {
+    pub liquidity_deployment: LiquidityDeployment,
+}
+
+#[cw_serde]
+pub struct RoundTrancheLiquidityDeploymentsResponse {
```
dusan-maksimovic, line 217: These structs should be added to `schema.rs` to get the JSON/TS schema generated.

contracts/hydro/src/contract.rs
```diff
@@ -1295,6 +1328,129 @@ fn withdraw_icq_funds(
...
+            "deployment",
+            serde_json_wasm::to_string(&deployment).map_err(|_| {
+                ContractError::Std(StdError::generic_err("Failed to serialize deployment"))
+            })?,
+        );
+
+    LIQUIDITY_DEPLOYMENTS_MAP.save(
+        deps.storage,
+        (round_id, tranche_id, proposal_id),
+        &deployment,
```
dusan-maksimovic, line 1419: Just checking: the inputs are not validated. Since the function can be called by the trusted 3rd party, maybe we don't need to validate them?

contracts/hydro/src/contract.rs
```diff
@@ -1295,6 +1328,129 @@ fn withdraw_icq_funds(
...
+        proposal_id,
+        destinations,
+        deployed_funds,
+        funds_before_deployment,
+        total_rounds,
+        remaining_rounds,
+    };
+
+    let response = Response::new()
+        .add_attribute("action", "add_round_liquidity_deployments")
```
dusan-maksimovic, line 1404: Let's align this with action name.

contracts/hydro/src/contract.rs
```diff
@@ -1295,6 +1328,129 @@ fn withdraw_icq_funds(
...
+    let constants = CONSTANTS.load(deps.storage)?;
+
+    validate_contract_is_not_paused(&constants)?;
+    validate_sender_is_whitelist_admin(&deps, &info)?;
+
+    // check that the deployment exists
+    LIQUIDITY_DEPLOYMENTS_MAP.load(deps.storage, (round_id, tranche_id, proposal_id))?;
+
+    let response = Response::new()
+        .add_attribute("action", "remove_round_liquidity_deployments")
```
dusan-maksimovic, line 1444: Same as above.

contracts/hydro/src/contract.rs
```diff
@@ -1295,6 +1328,129 @@ fn withdraw_icq_funds(
...
+// This function will add a given liquidity deployment to the deployments that were performed.
+// This will not actually perform any movement of funds; it is assumed to be called when
+// a trusted party, e.g. a multisig, has performed some deployment,
+// and the contract should be updated to reflect this.
+// This will return an error if:
+// * the given round has not started yet
+// * the given tranche does not exist
+// * the given proposal does not exist
+// * there already is a deployment for the given round, tranche, and proposal
+// Having more arguments rather than nested structures actually makes the json easier to construct when calling this from the outside,
```
dusan-maksimovic, line 1340: We could construct `LiquidityDeployment` from input parameters in the `execute()` function and pass it to `add_liquidity_deployment()`?

dusan-maksimovic (commented):
Left a few comments, otherwise looks great!
Just make sure to compile the contracts again and regenerate the schema.

contracts/hydro/src/contract.rs
p-offtermatt: nice idea, thanks!

dusan-maksimovic (approved):

PR #165, opened 2024-10-30
Custom bids duration

dusan-maksimovic:
## Description

Closes: #142 

This PR introduces a `bid_duration` field on the Proposal to specify for how many rounds the liquidity is requested. If user votes with a lock for a proposal spanning across multiple rounds, it will not be allowed to vote with the same lock in the next rounds until the round in which the liquidity is clawed back from the 'long lasting' proposal. Max bid duration is stored in `config` and can be updated by the whitelist admins.

---

### Author Checklist

I have...


contracts/hydro/src/contract.rs
```diff
@@ -715,71 +736,89 @@ fn vote(
         // Get any existing vote for this sender and reverse it- this may be a vote for a different proposal (if they are switching their vote),
         // or it may be a vote for the same proposal (if they have increased their power by locking more and want to update their vote).
         // TODO: this could be made more gas-efficient by using a separate path with fewer writes if the vote is for the same proposal
-        let vote = VOTE_MAP.load(
+        let vote = VOTE_MAP.may_load(
```
dusan-maksimovic, line 739: Apart from this load()->may-load() change, it's just an indentation change.
The None "arm" is newly added code.

.changelog/unreleased/features/165-custom-bid-durations.md
```diff
@@ -0,0 +1,2 @@
+- Custom bids duration.
```
p-offtermatt, line 1:
```suggestion
- Allow bids to set a custom duration they would like to receive liquidity for.
```

contracts/hydro/src/state.rs
```diff
@@ -48,10 +49,16 @@ pub struct Proposal {
     pub description: String,
     pub power: Uint128,
     pub percentage: Uint128,
+    pub bid_duration: u64, // number of rounds liquidity is allocated excluding voting round. default 1
```
p-offtermatt, line 52:
```suggestion
    pub bid_duration: u64, // number of rounds liquidity is allocated excluding voting round
```

the default is not actually enforced anywhere, right? I think it probably shouldn't mention the default here then

contracts/hydro/src/contract.rs
```diff
@@ -715,71 +736,89 @@ fn vote(
...
+                    deps.storage,
+                    ((round_id, tranche_id), info.sender.clone(), lock_id),
+                );
+
+                response = response.add_attribute(
+                    format!("lock_id_{}_old_proposal_id", lock_id),
+                    vote.prop_id.to_string(),
+                );
+            }
+            None => {
```
p-offtermatt, line 803: that's a smart way to do it, nice.

could you add a comment to the none branch, maybe, that this means that users can change their vote during a round, because we don't check this if users already voted?

contracts/hydro/src/contract.rs
```diff
@@ -715,71 +736,89 @@ fn vote(
...
+                // If user didn't yet vote with the given lock in the given round and tranche, check if they
+                // voted in previous rounds for some proposal that spans multiple rounds.
+                let voting_allowed_round = VOTING_ALLOWED_ROUND
+                    .may_load(deps.storage, (tranche_id, info.sender.clone(), lock_id))?;
+
+                if let Some(voting_allowed_round) = voting_allowed_round {
+                    if voting_allowed_round > round_id {
+                        return Err(ContractError::Std(
+                        StdError::generic_err(format!(
+                            "Not allowed to vote with lock_id {} on new proposal. Voted already on proposal still in bidding phase ending in round {}.",
```
p-offtermatt, line 813:
```suggestion
                            "Not allowed to vote with lock_id {} in tranche {}: Voted in round {} on proposal {} with duration {}; cannot vote again with this lock_id until round {}",
```
error should be a bit more descriptive imo

contracts/hydro/src/state.rs
```diff
@@ -48,10 +49,16 @@ pub struct Proposal {
...
 }
 
 // VOTE_MAP: key((round_id, tranche_id), sender_addr, lock_id) -> Vote
 pub const VOTE_MAP: Map<((u64, u64), Addr, u64), Vote> = Map::new("vote_map");
+
+// Tracks the next round in which user is allowed to vote with the given lock_id.
+// VOTING_ALLOWED_ROUND: key(tranche_id, sender_addr, lock_id) -> round_id
+pub const VOTING_ALLOWED_ROUND: Map<(u64, Addr, u64), u64> = Map::new("voting_allowed_round");
```
p-offtermatt, line 60: Do we need the sender_addr here? Not sure it helps to include it

contracts/hydro/src/contract.rs
```diff
@@ -831,7 +870,7 @@ fn vote(
             )?;
 
             // update the proposal in the proposal map, as well as the props by score map
-            update_proposal_and_props_by_score_maps(
+            let updated_proposal = update_proposal_and_props_by_score_maps(
```
p-offtermatt, line 873: I was a bit iffy on this (it's using the return for something completely different to what the func is doing), but I think it's elegant, since one can reasonably expect the func to have a return, so I think it ends up being quite nice :)

p-offtermatt (commented):
Very nice! I added a few comments.

Also, I think it would be good to have a test case to confirm that it is possible to switch ones vote during the round, even when voting for an extended-duration proposal. 

could easily be added as part of the test you added I think

contracts/hydro/src/contract.rs
dusan-maksimovic: Yeah, it got me thinking at first, but Bernd did it in a very smart way. :) 
Will add the comment.

contracts/hydro/src/contract.rs
dusan-maksimovic: Currently we can't extract the 'long lasting' proposal that user voted on. If you still think we should change this message, then we should store the round id in which voting occurred in the `VOTING_ALLOWED_ROUND`. WDYT, is it worth it?

contracts/hydro/src/contract.rs
dusan-maksimovic: Yeah, I had the same feeling about it. Good, we can go with this approach. :)

contracts/hydro/src/state.rs
dusan-maksimovic: We don't, I will remove it.

contracts/hydro/src/contract.rs
p-offtermatt: Let's just leave the proposal and round out and just say when they can vote again then

PR #194, opened 2024-12-02
Include lockup schedule as contract constant

p-offtermatt:
## Description

Closes: N/A

This is a refactor. This should make it easier for third parties to build on top of Hydro, since they will
not need to hardcode the lockup schedule, either.

---

### Author Checklist

I have...


contracts/hydro/src/msg.rs
```diff
@@ -20,8 +20,21 @@ pub struct InstantiateMsg {
...
-    pub is_in_pilot_mode: bool,
     pub max_deployment_duration: u64,
+    // A vector of tuples, where each tuple contains a round number and the power scaling factor
+    // that a lockup has when it has this many rounds left at the end of the round.
+    // The vector should be sorted by round number in ascending order.
+    // It will always be implicit that 0 rounds lock left corresponds to 0 voting power.
+    // Otherwise, it implicitly assumes that between two entries, the larger entries power is used.
+    // For example, if the vector is [(1, 1), (2, 1.25), (3, 1.5), (6, 2), (12, 4)],
+    // then the power scaling factors are
+    // 0x if lockup has expires before the end of the round
```
dusan-maksimovic, line 31: remove "has"

contracts/hydro/src/state.rs
```diff
@@ -17,8 +17,21 @@ pub struct Constants {
...
+    // Otherwise, it implicitly assumes that between two entries, the larger entries power is used.
+    // For example, if the vector is [(1, 1), (2, 1.25), (3, 1.5), (6, 2), (12, 4)],
+    // then the power scaling factors are
+    // 0x if lockup has expires before the end of the round
+    // 1x if lockup has between 0 and 1 epochs left at the end of the round
+    // 1.25x if lockup has between 1 and 2 epochs left at the end of the round
+    // 1.5x if lockup has between 2 and 3 epochs left at the end of the round
+    // 2x if lockup has between 3 and 6 epochs left at the end of the round
+    // 4x if lockup has between 6 and 12 epochs left at the end of the round
+    pub round_lock_power_schedule: Vec<(u64, Decimal)>,
```
dusan-maksimovic, line 34: WDYT about converting this Vec into a full-blown struct called RoundLockPowerSchedule? Then you could create a new() function that accepts list of tuples and validates them (e.g. no duplicates, must be provided in ascending order by remaining rounds number, etc.). This would be called from the instantiate() function. And you could also move the logic of validate_lock_duration() and scale_lockup_power() into separate methods of this new structure. It somehow feels weird to rely on these tuples being ordered in ascending order without actually validating that fact.

I would also convert this tuple into a structure with fields like "rounds_remaining" (not very good, please try to find a better one) and "power_scaling_factor", for better semantics.

dusan-maksimovic (approved):

PR #208, opened 2024-12-17
Add per tranche info for lockups

p-offtermatt:
## Description

Closes: N/A

Adds some information to the `query_all_user_lockups` function. This information is necessary to enable callers to know which lockups can currently vote, and which cannot.

---

### Author Checklist

I have...


contracts/hydro/src/contract.rs
```diff
@@ -1700,8 +1708,60 @@ pub fn query_all_user_lockups(
...
+                        // if the proposal voted in this round, we ignore the VOTING_ALLOWED_ROUND map,
+                        // since it just contains the future information on when the lockup will be able to vote again
+                        // if it doesn't change the vote, but it can vote in the current round either way
+                        current_round_id
+                    } else {
+                        // if the lockup has not voted in this round, VOTING_ALLOWED_ROUND does contain
+                        // current information on whether the votup can vote right now or not
+                        VOTING_ALLOWED_ROUND
+                            .may_load(deps.storage, (*tranche_id, lock.lock_entry.lock_id))
+                            .unwrap_or_default()
```
dusan-maksimovic, line 1742: I would prefer skipping entry if `may_load()` returned error (couldn't parse the value). This can be done by using `filter_map()` instead of `map()`.

contracts/hydro/src/contract.rs
```diff
@@ -1700,8 +1708,60 @@ pub fn query_all_user_lockups(
...
+                            .unwrap_or(0)
```
dusan-maksimovic, line 1743: If there is no entry in `VOTING_ALLOWED_ROUND` it could also mean it is a brand-new lockup, so I think we should return `current_round_id` instead of 0.

contracts/hydro/src/contract.rs
```diff
@@ -1700,8 +1708,60 @@ pub fn query_all_user_lockups(
...
+                    };
+
+                    // return the info for this tranche
+                    PerTrancheLockupInfo {
+                        tranche_id: *tranche_id,
+                        next_round_lockup_can_vote: next_round_voting_allowed,
+                        current_voted_on_proposal: voted_for_proposal,
```
dusan-maksimovic, line 1750: This is a bit misleading since it will only give info for which proposal the lockup voted for in the current round (if it voted). Maybe we are OK with that? Not sure is it worth to search for which proposal it voted in latest round before the current one.

contracts/hydro/src/contract.rs
```diff
@@ -1700,8 +1708,60 @@ pub fn query_all_user_lockups(
...
+                        .map(|vote| vote.prop_id);
+
+                    let next_round_voting_allowed: u64 = if voted_for_proposal.is_some() {
+                        // if the proposal voted in this round, we ignore the VOTING_ALLOWED_ROUND map,
+                        // since it just contains the future information on when the lockup will be able to vote again
+                        // if it doesn't change the vote, but it can vote in the current round either way
+                        current_round_id
+                    } else {
+                        // if the lockup has not voted in this round, VOTING_ALLOWED_ROUND does contain
+                        // current information on whether the votup can vote right now or not
```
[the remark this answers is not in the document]
p-offtermatt: oops D: what a typo!

contracts/hydro/src/contract.rs
p-offtermatt: is this comment in the wrong line? I already am doing unwrap_or_default here, same as above

contracts/hydro/src/contract.rs
p-offtermatt: ok, makes sense to me. anything before the current round is meant to just signify "it can vote now"

contracts/hydro/src/contract.rs
p-offtermatt: Yeah, this is just in the current round. I will add a comment. Doing this in general seems annoying with our state structure right now.

I will leave it as is for now, but it's one thing we might want to add later on

contracts/hydro/src/contract.rs
dusan-maksimovic: No, the line is correct. What I am saying is that `unwrap_or_default()` will return Some(0) if `may_load()` returned Err. I think we should skip that tranche instead.

p-offtermatt: Ok, I see, filter_map in the loop thats enclosing this. that makes sense to me, I adjusted it. please check it out :)

Issue #197, opened 2024-12-09
Allow passing lockup_ids for unlocking


p-offtermatt:
For vote aggregators, it's important that they can unlock only some locks, to avoid a user paying for unlocks on other users lockups.

We should consider whether to skip locks that are not actually expired, instead of erroring out.

arlai-mk:
Should this issue be closed?

p-offtermatt:
Yes! I think it was supposed to auto-close (from the PR) but maybe didn't because the PR came from a fork. I've closed manually

PR #213, opened 2025-01-06
Unlock specific lock_ids

arlai-mk:
## Description

Closes: *#197*

This PR allows passing `lock_ids` for unlocking. It is also backwards-compatible, in the way that if we don't pass `lock_ids`, it will have the same behavior as previously.

Also, it will skip locks that are not actually expired, instead of erroring out.

---

### Author Checklist

I have...


p-offtermatt (commented):
Awesome, lgtm! Could you add a simple changelog entry for it? E.g. see how it is done in https://github.com/informalsystems/hydro/blob/main/.changelog/v2.0.0/features/162-add-liquidity-deployments.md, you can similarly add one under .changelog/unreleased.

Other than that, looks good - I ran the interchain tests locally, and they still pass, so afaict this indeed is backward compatible for outside API calls

edit: clippy is also protesting something minor - I've adjusted that (you can check by running `make clippy` locally)
Normally it runs in the CI, but I think something doesn't work because it's from an outside fork, I'll check on that

p-offtermatt (approved):

p-offtermatt:
Alright, looking good! I will merge this in!

Issue #214, opened 2025-01-07
Allow querying a specified list of locks for a user


arlai-mk:
Vote aggregators, such as Zephyrus, will likely manage a significant number of user locks within the Hydro platform. Currently, retrieving lock information requires using `QueryMsg::AllUserLockups`, which can be inefficient and time-consuming for aggregators with a large number of locks.

To improve performance and reduce query latency, it could be good to introduce a new query message: `QueryMsg::SomeUserLockups` (name can be different).

Suggested `QueryMsg::SomeUserLockups` parameters:
- `address`: The user's (or, in the above case, aggregator's smart contract) address.
- `lock_ids`: A vector containing the specific `lock_ids` for which we want to retrieve lock information.

p-offtermatt:
Great idea!

p-offtermatt:
in general, I think it's easiest to skip. The caller will be able to tell which ones are missing by looking at the list of returns

PR #216, opened 2025-01-17
Adding SomeUserLockups and SomeUserLockupsWithTrancheInfos queries

arlai-mk:
## Description

Closes: *#214*

Vote aggregators, such as Zephyrus, will likely manage a significant number of user locks within the Hydro platform. Currently, retrieving lock information requires using `QueryMsg::AllUserLockups` or the new `QueryMsg::AllUserLockupsWithTrancheInfos`, which can be inefficient and time-consuming for aggregators with a large number of locks.

This PR introduces 2 new queries: `QueryMsg::SomeUserLockups` and `QueryMsg::SomeUserLockupsWithTrancheInfos`, which will not have a `start_from` or `limit` parameter, but instead will pass a list of IDs which are the only ones that might be returned (if they correctly belong to the specified `address` (in the case of vote aggregators, the `address` would be the vote aggregators contract address).

In this PR, I created two new `SomeUserLockupsResponse` and `SomeUserLockupsWithTrancheInfosResponse` structs, even if they are exactly the same as the existing `AllUserLockupsResponse` and `AllUserLockupsWithTrancheInfosResponse` structs, because codegen seemed to force the query response to be of that type, even if I informed that they should return the same existing structs.

Even defining `pub type SomeUserLockupsResponse = AllUserLockupsResponse;` did not seem to work with `ts-codegen`. If you find a solution to just have the same type as Response for both `All` and `Some` queries, that would probably be cleaner.

---

### Author Checklist

I have...


contracts/hydro/src/query.rs
```diff
@@ -25,6 +25,9 @@ pub enum QueryMsg {
         limit: u32,
     },
 
+    #[returns(SomeUserLockupsResponse)]
+    SomeUserLockups { address: String, lock_ids: Vec<u64> },
```
dusan-maksimovic, line 29: `SpecificUserLockups` sounds better to me. WDYT?

contracts/hydro/src/query.rs
```diff
@@ -160,12 +166,26 @@ pub struct AllUserLockupsResponse {
     pub lockups: Vec<LockEntryWithPower>,
 }
 
+// This is necessary because otherwise, cosmwasm-ts-codegen does not generate SomeUserLockupsResponse
```
dusan-maksimovic, line 169: Yeah, I've run into this issue before, tried different things, eventually did the same thing you did. Just leave it be, don't worry about it.

contracts/hydro/src/query.rs
arlai-mk: I'm fine with either name. I'm happy to go with your suggestion of `SpecificUserLockups` if you feel it's more descriptive.

dusan-maksimovic: Let's go with `SpecificUserLockups`, sounds a bit more descriptive. Same for the other one. Thanks!

contracts/hydro/src/query.rs
arlai-mk: Done

contracts/hydro/src/query.rs
arlai-mk: Done

contracts/hydro/src/testing_queries.rs
```diff
@@ -3,7 +3,8 @@ use std::str::FromStr;
 
 use crate::contract::{
     compute_current_round_id, query_all_user_lockups, query_all_user_lockups_with_tranche_infos,
-    query_user_votes, scale_lockup_power,
+    query_some_user_lockups, query_some_user_lockups_with_tranche_infos, query_user_votes,
```
dusan-maksimovic, line 6: Let's also rename these functions to "specific" instead of "some".

dusan-maksimovic (approved):
Just this one last minor thing with function names, otherwise LGTM!
Thanks for your contribution!

contracts/hydro/src/testing_queries.rs
arlai-mk: Sorry, missed changing those names. It's done.

dusan-maksimovic: Please recompile the contracts as well. I know it's a bit tiring, but those .wasm files are what we use in e2e tests, etc.

arlai-mk: I actually did recompile, but changing the internal function names doesn't seem to affect the .wasm files

dusan-maksimovic: Oh, OK, didn't expect that! Then we are good to merge!

PR #220, opened 2025-01-23
Introduce compounder cap

dusan-maksimovic:
## Description

This PR introduces the "compounder cap" which reserves certain amount of tokens (out of the total number of tokens allowed to be locked) and allows users that had voting power in previous round to exclusively lock in this cap, for a certain time frame. The "compounder cap" is meant to be active for some time at the beginning of the round. After this time has expired, any amount of tokens that wasn't locked by the compunders will be at disposal to any other user to freely lock, up to the limit of total cap, that is enforced at any time. Implementation also supports re-activation of "compunders cap" at any time during the round, which is controlled by the contract governance and handled by the `update_config()` action (as it is the case with initial activation).

### Techinical Details
Updated contract stores:
- CONSTANTS are converted into Map, to allow different configurations activation depending on the time. This allows us to e.g. prepare 2 configurations- one that increases total_cap and activates the compunder cap, and the other one that is activated some time after and sets the compounder cap to zero.
- LOCKS_MAP converted to SnapshotMap so that we can compute users voting power at a given height. This goes along with newly added storage USER_LOCKS which allows us to know which lockups user had at a given height, since we can not iterate over LOCKS_MAP at desired height in the past.

Newly added stores:
- EXTRA_LOCKED_TOKENS_ROUND_TOTAL and EXTRA_LOCKED_TOKENS_CURRENT_USERS allow us to track how many tokens were locked in compunder cap and enforce the maximum limit per user and in total.
- TOTAL_VOTING_POWER_PER_ROUND stores total voting power for current and all future rounds in which there is at least some voting power. It gets updated each time user locks/relocks as well as on validator power ratio changes. This store will be more useful for governance but it is introduced together with user voting power snapshot map. For compunder cap PR, we could compute this value for past rounds on the fly since we store locked shares and validator power ratios per round, but this is also a bit more performant.
- ROUND_TO_HEIGHT_RANGE and HEIGHT_TO_ROUND are introduced to be able to this conversion, where for the compounder cap we need round->height, while for the governance we will need the other direction. Both stores are updated on each transaction that gets executed against the smart contract. The function for getting round for height will be implemented in governance PR.

---

### Author Checklist

I have...

* [ ] Targeted the correct branch
* [ ] Included the necessary unit tests


contracts/hydro/src/lsm_integration.rs
```diff
@@ -22,16 +21,15 @@ pub const COSMOS_VALIDATOR_ADDR_LENGTH: usize = 52; // e.g. cosmosvaloper15w6ra6
...
 // Returns OK if the denom is a valid IBC denom representing LSM
 // tokenized share transferred directly from the Cosmos Hub
-// of a validator that is also currently among the top
-// max_validators validators, and returns the address of that validator.
+// of a validator that is also among the top max_validators validators
+// for the given round, and returns the address of that validator.
 pub fn validate_denom(
     deps: Deps<NeutronQuery>,
-    env: Env,
+    round_id: u64,
```
dusan-maksimovic, line 28: This is to support denom resolution for the past rounds. We need to pass in the round_id because sometimes this function gets executed in current round, but we are interested for past rounds (e.g. for compounder cap to determine users voting power in previous round).

contracts/hydro/src/state.rs
```diff
@@ -58,7 +59,15 @@ pub struct Constants {
     pub round_length: u64,
     pub lock_epoch_length: u64,
     pub first_round_start: Timestamp,
+    // The maximum number of tokens that can be locked by any users (currently known and the future ones)
     pub max_locked_tokens: u128,
+    // The maximum number of tokens (out of the max_locked_tokens) that is reserved for locking only
+    // for currently known users. This field is intended to be set to some value greater than zero at
+    // the begining of the round, and such Constants would apply only for predefined period of time.
```
p-offtermatt, line 66:
```suggestion
    // the begining of the round, and such Constants would apply only for a predefined period of time.
```

p-offtermatt (commented):
Reviewed: 
* Time-adjustment for constants
* Keeping historical values for user locks and total power

To-do:
* Compounder cap itself

contracts/hydro/src/state.rs
```diff
@@ -228,3 +272,24 @@ impl ValidatorInfo {
...
+// contract in the given round.
+// Notice that the round could span beyond these boundaries, but we don't have a way to know that.
+// Besides, the info we store here is sufficient for our needs.
+// ROUND_TO_HEIGHT_RANGE: key(round_id) -> HeightRange
+pub const ROUND_TO_HEIGHT_RANGE: Map<u64, HeightRange> = Map::new("round_to_height_range");
+
+// Stores the mapping between the block height and round. It gets populated
+// each time a transaction is executed against the smart contract.
+// HEIGHT_TO_ROUND: key(block_height) -> round_id
+pub const HEIGHT_TO_ROUND: Map<u64, u64> = Map::new("height_to_round");
```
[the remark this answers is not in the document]
dusan-maksimovic: Due to the nature of the SnapshotItem that, when queried at a height on which change occurred it returns old value, @p-offtermatt and I decided to stick to old approach.

p-offtermatt (commented):
Just to check my understanding of the migration:
* USER_LOCKS will only be populated from the height of the migration onwards, and empty for earlier rounds (thus get_user_voting_power_for_past_rounds will be unpopulated)
* TOTAL_VOTING_POWER_PER_ROUND will be populated correctly for all rounds
* round height mappings will be populated only for future heights, and empty for earlier rounds


Do you think it's worth storing the height of the migration and erroring out when querying these if the asked-for height was before the migration height?

Pro: harder to mess up when calling these
Con: might make testing harder

Maybe we can do this just in the outward-facing queries. I think that would not impact testing too much, but assure us a bit more that noone will build on this and have it silently fail in a weird way.

dusan-maksimovic:
> Just to check my understanding of the migration:
> 
> * USER_LOCKS will only be populated from the height of the migration onwards, and empty for earlier rounds (thus get_user_voting_power_for_past_rounds will be unpopulated)
> * TOTAL_VOTING_POWER_PER_ROUND will be populated correctly for all rounds
> * round height mappings will be populated only for future heights, and empty for earlier rounds

That is correct.
1. Just a clarification on this item, get_user_voting_power_for_past_round() will be used in compounder cap and governance. For compounder cap, we will migrate the contract before e.g. the round 4 is finished, and starting from round 5 we can introduce the compounder cap. So we will not need it for past rounds. For the governance, we only care to have this info at the height first proposal is submitted, which will be in the future, so we are good there as well.

I agree with storing the info during the migration and checking it later. Currently we only need this for the mentioned function, even though it doesn't get called from the outside (yet).

contracts/hydro/src/utils.rs
```diff
@@ -0,0 +1,290 @@
...
+    let total_locked_after = total_locked_tokens + amount_to_lock;
+    if total_locked_after > constants.max_locked_tokens {
+        return lock_limit_reached_error;
+    }
+
+    let public_cap = constants.max_locked_tokens - constants.current_users_extra_cap;
+
+    // This branch will be executed in one of the following cases:
+    //   1) constants.current_users_extra_cap != 0 and there is SOME room in the public cap
+    //   2) constants.current_users_extra_cap == 0 and there is ENOUGH room in the public_cap.
```
p-offtermatt, line 87: why do you list the special case here? isn't this branch just always executed if there is *some* room in the public cap, and then the check public_cap > total_locked_after handles the case where there is enough room in the public cap?

contracts/hydro/src/utils.rs
```diff
@@ -0,0 +1,290 @@
...
+        if public_cap >= total_locked_after {
+            return Ok(LockingInfo {
+                lock_in_public_cap: Some(amount_to_lock),
+                lock_in_known_users_cap: None,
+            });
+        }
+
+        // Lock as much as possible in the public_cap, and the rest in the extra_cap
+        let lock_in_public_cap = public_cap - total_locked_tokens;
+        let lock_in_known_users_cap = amount_to_lock - lock_in_public_cap;
```
p-offtermatt, line 101: Suggest to make the naming consistent and call it either known_users_cap or _extra_cap everywhere.
probably, calling it known_users_cap seems better, actually!

contracts/hydro/src/utils.rs
```diff
@@ -0,0 +1,290 @@
...
+        .unwrap_or(0);
+
+    if extra_locked_tokens_round_total + amount_to_lock > constants.current_users_extra_cap {
+        return Ok(false);
+    }
+
+    // Determine if user has the right to lock in extra_cap by looking at its voting power in previous round.
+    // If we are in round 0 then check the current round.
+    let round_to_check = match current_round {
+        0 => current_round,
```
p-offtermatt, line 160: should we just return false in this case, maybe?

contracts/hydro/src/utils.rs
```diff
@@ -0,0 +1,290 @@
...
+
+    // Calculate what would be users share of extra locked tokens in the maximum allowed extra locked tokens
+    let users_current_extra_lock = EXTRA_LOCKED_TOKENS_CURRENT_USERS
+        .may_load(deps.storage, (current_round, sender.clone()))?
+        .unwrap_or(0);
+
+    let users_extra_lock =
+        Decimal::from_ratio(users_current_extra_lock + amount_to_lock, Uint128::one());
+    let maximum_extra_lock = Decimal::from_ratio(constants.current_users_extra_cap, Uint128::one());
+    let users_extra_lock_share = users_extra_lock.checked_div(maximum_extra_lock)?;
```
p-offtermatt, line 191: this should also have a check around maximum_extra_lock to avoid division by zero, maybe? I don't think this can happen in the current code, but I feel it would benefit from being made safe from division by zero in case we change the code

p-offtermatt (commented):
Reviewed everything except the tests for now.
Great work!

contracts/hydro/src/utils.rs
dusan-maksimovic: Because if you look at the code, you might wonder what happens if `public_cap >= total_locked_after` gives `false` and you try to lock in extra_cap while it is not active (`constants.current_users_extra_cap == 0`), which shouldn't be allowed (even though there is also a check for this in can_user_lock_in_extra_cap()). It is just a clarification since the upper condition in that case becomes ` if total_cap >= total_locked_after`, which must always be true, otherwise we wouldn't be executing that line at all. Anyways, it's just something I thought could be useful to document, but I can remove it if you disagree.

p-offtermatt: I found the comment a bit confusing, but let's leave it if you think it's helpful

contracts/hydro/src/testing_compounder_cap.rs
```diff
@@ -0,0 +1,555 @@
...
+// 7.  Round 1: Have the known user from previous step lock more to fill the public_cap and some more into known_users_cap.
+// 8.  Round 1: Have the same known user try to lock in known_users_cap more than it should be allowed.
+// 9.  Round 1: Have the same known user lock the most it should be allowed in the known_users_cap.
+// 10. Round 1: Have other two known users lock as much as they should be allowed in the known_users_cap.
+// 11. Round 1: Update config to increase total_cap and set known_users_cap starting from round 2.
+// 12. Round 1: Update config to close the known_users_cap after some time in round 2.
+// 13. Round 2: Have a completely new user lock tokens to fill up the public_cap, then try to lock more.
+// 14. Round 2: Have a known user lock maximum allowed in known users cap.
+// 15. Round 2: Advance the chain to end the known_users_cap duration and have a user from step #13 lock
+//     additional amount that matches the entire amount previously reserved for known_users_cap.
```
p-offtermatt, line 57: In principle, I think I would have preferred a more focussed table-based test here, e.g.:
Try to have a user lock up:
* the public cap is large enough -> works and their known user cap remains unfilled
* public cap is not large enough, but known user cap is -> works, and their known user cap is partially filled now
* total is not large enough -> fails
* total would be large enough, but users known user cap is too small -> fails

I think we can leave this as-is, but just in general I think the long tests that try to test many things in one test feel less maintainable (for sure this is not the only place in the codebase where we have tests like this though).

contracts/hydro/src/testing_lsm_integration.rs
```diff
@@ -736,7 +740,30 @@ fn lock_tokens_multiple_validators_and_vote() {
     }
 
     // update the power ratio for validator 1 to become 0.5
-    set_validator_power_ratio(deps.as_mut().storage, 0, VALIDATOR_1, Decimal::percent(50));
+    let mock_tokens = Uint128::new(500);
```
p-offtermatt, line 743: could this be a util function somewhere? I know it's not really used anywhere else right now, but it seems like the type of code block that should be hidden in a function :)

p-offtermatt:
Ok, LGTM! Just a reminder to do the admin things (recompile & add changelog). We can do a separate PR for the queries.

contracts/hydro/src/testing_lsm_integration.rs
dusan-maksimovic: Not sure what do you mean by that. `get_mock_validator()` is already a util function, but we want to keep it flexible, so the tokens and shares are provided by the caller. We could add a `shares_from_tokens()`, perhaps this is what you meant?

p-offtermatt: the whole block up to the sudo call is what I meant

dusan-maksimovic: Done in https://github.com/informalsystems/hydro/pull/220/commits/400c1abb6ebaa9c98891b9ac0a71b262c6aea30e

p-offtermatt (approved):

PR #231, opened 2025-02-25
Allow voting with locks whose voted-for-proposal did not receive any funds

p-offtermatt:
## Description

Closes: N/A

This makes it so that if a lock has voted for a long-deployment-duration proposal in a past round,
but that proposal did not end up getting any funds deployed, then the lock can vote again.

This adds:
* logic in process_votes to let locks vote
* logic in enrich_lockups_with_tranche_infos to properly reflect this in the query endpoints
* a unit test to verify the new behaviour

---

### Author Checklist

I have...

### DIFF
--- a/contracts/hydro/src/contract.rs
+++ b/contracts/hydro/src/contract.rs
@@ -1,4 +1,5 @@
 use std::collections::{HashMap, HashSet};
+use std::convert;
 
 // entry_point is being used but for some reason clippy doesn't see that, hence the allow attribute here
 #[allow(unused_imports)]
@@ -43,9 +44,9 @@ use crate::state::{
     VALIDATOR_TO_QUERY_ID, VOTE_MAP, VOTING_ALLOWED_ROUND, WHITELIST, WHITELIST_ADMINS,
 };
 use crate::utils::{
-    get_current_user_voting_power, get_lock_time_weighted_shares,
-    load_constants_active_at_timestamp, load_current_constants, run_on_each_transaction,
-    scale_lockup_power, to_lockup_with_power, update_locked_tokens_info,
+    find_deployment_for_voted_lock, get_current_user_voting_power, get_lock_time_weighted_shares,
+    has_nonzero_funds, load_constants_active_at_timestamp, load_current_constants,
+    run_on_each_transaction, scale_lockup_power, to_lockup_with_power, update_locked_tokens_info,
     validate_locked_tokens_caps,
 };
 use crate::validators_icqs::{
@@ -1814,6 +1815,7 @@ fn enrich_lockups_with_tranche_infos(
                         } else {
                             // if the lockup has not voted in this round, VOTING_ALLOWED_ROUND does contain
                             // current information on whether the lockup can vote right now or not
+
                             VOTING_ALLOWED_ROUND
                                 .may_load(deps.storage, (*tranche_id, lock.lock_entry.lock_id))
                                 .map(|voting_allowed_round| {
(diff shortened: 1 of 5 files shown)


contracts/hydro/src/contract.rs
```diff
@@ -1827,7 +1827,27 @@ fn enrich_lockups_with_tranche_infos(
...
+                    // check whether there is a deployment associated with that proposal
+                    if next_round_voting_allowed > current_round_id {
+                        let deployment = find_deployment_for_voted_lock(
+                            deps,
+                            current_round_id,
+                            *tranche_id,
+                            &converted_addr,
+                            lock.lock_entry.lock_id,
+                        )
+                        .expect("Failed to find deployment for voted lock");
```
dusan-maksimovic, line 1843: Even though we shouldn't encounter `Err` here, let's replace the `expect()` call with returning None instead (since this chunk of code is within `filter_map()` call. This way our query wouldn't panic in case of unexpected errors. Basically do the same thing as we do when handling all other `Result<>` in this function.

contracts/hydro/src/utils.rs
```diff
@@ -543,3 +545,62 @@ pub struct LockingInfo {
     pub lock_in_public_cap: Option<u128>,
     pub lock_in_known_users_cap: Option<u128>,
 }
+
+pub fn has_nonzero_funds(liquidity_deployment: LiquidityDeployment) -> bool {
```
dusan-maksimovic, line 549: Nitpick: this could be implemented as a method of the `LiquidityDeployment` struct.

contracts/hydro/src/utils.rs
```diff
@@ -543,3 +545,62 @@ pub struct LockingInfo {
...
+// It will return an error if the lock has not voted for any proposal,
+// or if the store entry for the proposals deployment cannot be parsed.
+pub fn find_deployment_for_voted_lock(
+    deps: &Deps<NeutronQuery>,
+    current_round_id: u64,
+    tranche_id: u64,
+    lock_voter: &Addr,
+    lock_id: u64,
+) -> Result<Option<LiquidityDeployment>, ContractError> {
+    let mut check_round = current_round_id - 1;
```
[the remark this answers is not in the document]
p-offtermatt: Done, thanks!

contracts/hydro/src/contract.rs
p-offtermatt: feels like it would be weird to just put none here when really something in the store is broken, no?

contracts/hydro/src/utils.rs
```diff
@@ -543,3 +545,62 @@ pub struct LockingInfo {
...
+                    (
+                        prev_proposal.round_id,
+                        prev_proposal.tranche_id,
+                        prev_proposal.proposal_id,
+                    ),
+                )
+                .map_err(|_| {
+                    // if we cannot read the store, there is an error
+                    ContractError::Std(StdError::generic_err(format!(
+                        "Could not read deployment store for proposal {} in tranche {}.",
```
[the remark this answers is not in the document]
p-offtermatt: done!

contracts/hydro/src/contract.rs
dusan-maksimovic: True, but we are doing it in two other cases (if we can't parse `VOTE_MAP` or `VOTING_ALLOWED_ROUND`) so this would make things consistent. On top of that, we would remove `expect()` call which shouldn't be used in production code because it can panic.

dusan-maksimovic (approved):

PR #235, opened 2025-03-03
Add information about which proposal lockups are tied to

p-offtermatt:
## Description

Closes: 

Adds to the existing enrich_lockups function some logic to fill in the proposal to which a lockup is tied, if it is indeed tied to any.

---

### Author Checklist

I have...


contracts/hydro/src/query.rs
```diff
@@ -154,6 +154,10 @@ pub struct PerTrancheLockupInfo {
     // In particular, if the lockup is blocked from voting in the current round (because it voted for a
     // proposal with a long deployment duration in a previous round), this will be None.
     pub current_voted_on_proposal: Option<u64>,
+
+    // This is the id of the proposal that the lockup is tied to because it has voted for a proposal with a long deployment duration.
+    // In case the proposal can currently vote (and is not tied to a proposal), this will be None.
```
dusan-maksimovic, line 159: Should this comment mention that also if the proposal that the lock voted on didn't receive any funds, this field will have value None?

Also in second sentence it should be "lock" instead of "proposal".

dusan-maksimovic (approved):
LGTM!

contracts/hydro/src/query.rs
p-offtermatt: I think that should be clear by the fact that it returns none if the lockup can be used to vote currently, but I will add a sentence

thx for noticing the typo, good call

PR #238, opened 2025-03-21
Added support for locking additional token groups

dusan-maksimovic:
## Description

This PR adds support for locking different tokens (mainly the base token derivatives) in addition to previously supported LSMs. The main changes are:
- Users will be able to lock LSMs and stDERIVATIVE tokens.
- On Hydro SC instantiation, one or more `Token Info Provider` instantiation messages are specified. If one of those messages is for the stDERIVATIVE token info provider, a new SC is instantiated during Hydro SC instantiation. In the `reply()` message, address of newly instantiated SC is saved in Hydro so that it can be used later when needed. If the LSM one is provided, a proper data structure is initialized in the Hydro SC.
- New smart contract called `ST Token Info Provider` is introduced. Its InstantiateMsg contains token denom representing the wanted ST derivative (e.g. stATOM) and the token group ID (e.g. statom). This SC is called by the Hydro SC when needed to validate if the provided denoms are expected ST derivative, and to obtain its ratio towards the base denom for the given round. Contract is responsible for tracking ratios of stDERIVATIVE to base token for current and previous rounds. This contract is also allow-listed on the Hydro SC to update the token power ratio in Hydro, when the ratio gets updated in the `ST Token Info Provider` SC itself. For this first version, `ST Token Info Provider` will return power ratio 1 for any given round.
- LSM Token Info Provider currently isn't extracted into a separate SC, since this would require some time to migrate the existing SC and its ICQs together with history into a separate SC, which is not worth the effort at the moment. At some point, this will be done.
- A new package called `token-info-provider-interface` is added to the Hydro workspace. It will contain messages, queries and query responses that will be shared between Hydro and Token Info Provider SCs.
- `TokenManager` is a new concept in Hydro SC that handles denom resolving and obtaining the power ratio of the given denom towards the base denom. The structure wraps around all Token Info Providers known to the given Hydro SC instance. The structure also implements caching, so for example, if user votes with multiple lockups containing stATOM, this will not cause `ST Token Info Provider` SC instance to be called multiple times, thus saving gas on execution. Cache is maintained during the transaction/query scope.
- Functions moved from `lsm_integration.rs` into `score_keeper.rs` since they belong there better:
        get_total_power_for_round()
	update_stores_due_to_power_ratio_change()
	update_scores_due_to_power_ratio_change()
	update_total_power_due_to_power_ratio_change()
	get_validator_shares_for_round() -> get_token_group_shares_for_round()
	add_validator_shares_to_round_total() -> add_token_group_shares_to_round_total()
- Moved lock&unlock tests into separate file `testing_locking_unlocking.rs` (because the `testing.rs` is becoming way too large)
---

### Author Checklist

I have...


contracts/hydro/src/msg.rs
```diff
@@ -149,3 +214,8 @@ pub struct LiquidityDeployment {
     // if this is a "repeating" proposal
     pub remaining_rounds: u64,
 }
+
+#[derive(Serialize, Deserialize)]
+pub enum ReplyPayload {
```
dusan-maksimovic, line 219: I added a new `ReplyPayload` structure because I didn't want to mix it up with the existing one that is specific to the ICQs and will be moved into a different SC at some point.

contracts/hydro/src/contract.rs
```diff
@@ -1497,6 +1551,118 @@ pub fn remove_liquidity_deployment(
...
+        .iter()
+        .map(|token_group| TokenGroupRatioChange {
+            token_group_id: token_group.0.clone(),
+            old_ratio: *token_group.1,
+            new_ratio: Decimal::zero(),
+        })
+        .collect();
+
+    // Remove any voting power on proposals and rounds that comes from tokens of the given token info provider.
+    apply_token_groups_ratio_changes(
```
dusan-maksimovic, line 1651: If we are removing LSM token info provider, this can be very gas consuming and potentially too large tx to enter the block. The only mitigation that we have at this moment is to perform the removal at the beginning of the round when there are no proposals. Updating just the rounds total power to remove all validators token groups should work.

One way to handle this properly would be to only store the token groups that need to be removed and introduce a new permission-less action to remove them in chunks. Since we are unlikely to remove LSM token info provider in the near future, I wouldn't do this part just yet.

contracts/hydro/src/contract.rs
```diff
@@ -487,38 +515,44 @@ fn refresh_single_lock(
...
-        return Err(ContractError::Std(StdError::generic_err(
-            "Lock denom is for a validator who is currently not in the set, try refreshing when the validator has enoug delegation",
-        )));
-    }
-    let validator = validator_result.unwrap();
+
+    let token_group_id = match validate_denom_result {
+        Ok(token_group_id) => token_group_id,
+        Err(err) => {
+            return Err(new_generic_error(format!(
```
p-offtermatt, line 527: Nit: nested errors should not repeat that they are an error
(since it will lead to things like e.g. "Can't refresh lock. Error: Can't validate denom. Error: Can't find channel 468")

p-offtermatt: I would take inspiration from https://github.com/informalsystems/hydro/blob/dusan/locking-different-tokens/contracts/hydro/src/contract.rs#L325 and have it be "validating denom: {}", error

contracts/hydro/src/contract.rs
```diff
@@ -1223,6 +1269,14 @@ fn create_icqs_for_validators(
 ) -> Result<Response<NeutronMsg>, ContractError> {
     let constants = load_current_constants(&deps.as_ref(), &env)?;
     validate_contract_is_not_paused(&constants)?;
+
+    let lsm_token_info_provider =
+        match TokenManager::new(&deps.as_ref()).get_lsm_token_info_provider() {
+            None => return Err(new_generic_error(
+                "Can not create validator ICQs- contract doesn't support locking of LSM tokens.",
```
p-offtermatt, line 1276:
```suggestion
                "Can not create validator ICQs - contract doesn't support locking of LSM tokens.",
```

contracts/hydro/src/contract.rs
p-offtermatt: Thanks, good idea. Neutron is about to raise gas ~10x with their Mercury upgrade, so maybe they will make our life much easier :D

contracts/hydro/src/token_manager.rs
```diff
@@ -0,0 +1,441 @@
...
+        let mut errors = vec![];
+
+        for provider in self.token_info_providers.iter_mut() {
+            match provider.resolve_denom(deps, round_id, denom.clone()) {
+                Ok(denom_group_id) => return Ok(denom_group_id),
+                Err(err) => errors.push(err),
+            }
+        }
+
+        // If there is only one token information provider, return the specific error.
```
p-offtermatt, line 65: very nice error handling here!

contracts/hydro/src/testing.rs
```diff
@@ -3763,3 +3292,218 @@ fn test_cannot_vote_while_long_deployment_ongoing() {
...
+        assert!(res.is_ok());
+    }
+
+    verify_proposals_and_rounds_powers(
+        &deps.storage,
+        first_round_id,
+        tranche_id,
+        &[(first_proposal_id, 0), (second_proposal_id, 0)],
+        &[(first_round_id, 0), (second_round_id, 0)],
+    );
```
p-offtermatt, line 3467: I think we should test somewhere that its possible to unlock tokens after the respective token_info_provider was removed

p-offtermatt: commented here because it seems like the easiest place to check that

p-offtermatt:
I wonder if we need to show explicitly which lockups are effectively "unusable" when a token_info_provider is removed. right now, you can just tell by seeing that their power is 0; but for expired lockups, it will be really hard to know which one is just expired and can be refreshed, and which cannot be refreshed because the token_info_provider was removed.

I'm not entirely sure what is the best idea here, my first instinct would be to essentially make validate_denom(denom) accessible as a query. the frontend would query that, and for lockups where the returned ratio is 0, it can show that this lockup is out.

This is pretty close to the validator power ratio query I added that we ended up removing, so I guess this is what I had in mind with that, but we never ended up having that on the frontend

contracts/hydro/src/testing_token_manager.rs
```diff
@@ -0,0 +1,405 @@
...
+            panic!("expected derivative token info provider, found LSM one.")
+        }
+        TokenInfoProvider::Derivative(provider) => {
+            assert_eq!(provider.contract, contract_address);
+        }
+    }
+}
+
+#[test]
+fn add_remove_token_info_provider_test() {
```
p-offtermatt, line 249: for all the new test cases I would encourage having brief summary comments above

p-offtermatt (commented):
also, if a token_info_provider is removed, all the locks

contracts/hydro/src/testing.rs
dusan-maksimovic: > I think we should test somewhere that its possible to unlock tokens after the respective token_info_provider was removed

We don't allow this currently, even before token info providers were introduced. It's the same with LSTs, e.g. if some validator drops out of the top N, users still wouldn't be able to unlock their tokens until their lockup expires.
I agree we should add this functionality. But also check if user voted for some longer proposal deployment and don't allow unlock in that case. Since it is not entirely related to this PR, I would do this in a separate one. Would you agree?

dusan-maksimovic:
> Maybe there could be a test of what happens when you re-add a token_info_provider that was removed? the tokens that are still locked should just re-gain voting power, right?

Will add a test case. That is correct, but in addition to that, I will need to add some code that goes through all proposals and rounds that "hold" tokens that belong to re-added token info providers and update their powers from 0 to some value. I didn't think this use case is very realistic so I didn't cover it, but will do it now.

contracts/hydro/src/testing.rs
p-offtermatt: Just to clarify: I am not talking about immediately unlocking. I am just talking about being able to unlock at all when the lockup expires

p-offtermatt:
> > Maybe there could be a test of what happens when you re-add a token_info_provider that was removed? the tokens that are still locked should just re-gain voting power, right?
> 
> Will add a test case. That is correct, but in addition to that, I will need to add some code that goes through all proposals and rounds that "hold" tokens that belong to re-added token info providers and update their powers from 0 to some value. I didn't think this use case is very realistic so I didn't cover it, but will do it now.

I am not sure whether it's super important to add this behaviour, unless it's very easy; it doesn't seem super realistic indeed, but we for sure should document that behaviour I think

contracts/hydro/src/testing.rs
dusan-maksimovic: We don't have problem with that, since we don't do denom validation on token unlocks. Just collect all the expired ones and send them back to the user.

PR #243, opened 2025-04-17
Gatekeeper contract implementation

dusan-maksimovic:
## Description

This PR brings the following changes:

- Introduction of the Gatekeeper smart contract, which is responsible for validating if the given user should be allowed to lock the given amount of tokens at a given point in time.
- Modifies the Hydro contract to be able to utilize the Gatekeeper. If specified in the Hydro instantiate message, Gatekeeper will be instantiated as well. The LockTokens() action is modified so that users can provide the merkle proofs that they should be allowed to lock the given number of tokens in Hydro.
- Renames "token-info-provider-interface" package into just "interface" and will use this package in the future to put there all types (basically DTOs) shared between Hydro workspace contracts. This allows us to have unified interface between our contracts without the need to reference each other.
	- existing "interface" package has been renamed to "cw-orch-interface"
- Changes the way schemas are generated- now we don't need to add each query type into schema.rs separately

---

### Author Checklist

I have...


contracts/gatekeeper/src/contract.rs
```diff
@@ -0,0 +1,416 @@
...
+            Some(Bound::exclusive(activate_at.nanos())),
+            None,
+            Order::Ascending,
+        )
+        .any(|_| true)
+    {
+        return Err(new_generic_error(
+            "Cannot register new stage at a timestamp earlier than already existing stage.",
+        ));
+    }
```
p-offtermatt, line 125: why this condition? can't we just allow arbitrary modifications (potentially excluding modifying a stage before the current time)

p-offtermatt (commented):
Linking the Hydro contract with the gatekeeper contract can only happen during instantiation afaict. This probably needs to also allow setting/removing the gatekeeper by a whitelist admin afterwards, otherwise we'd have to migrate the contract to change the gatekeeper

contracts/gatekeeper/src/contract.rs
p-offtermatt: Left it like this, doesn't seem too problematic in the end, it just means we should not put too many stages ahead of time (since it makes it more likely some early stage could need corrections)

p-offtermatt (approved):

p-offtermatt:
Doesn't seem like the interchain tests are/were working on this one, trying to investigate that

contracts/gatekeeper/src/contract.rs
dusan-maksimovic: The issue is with how we determine the stage-epoch relation and how many tokens users locked per given epoch. If we allowed new stages to be created at timestamps earlier than the already existing ones, it could lead to situations like the following:
Stage ID            Stage start            New Epoch
1                         5.5.2025 10 AM    Yes
2                         5.5.2025 04 PM    Yes
3                         5.5.2025 02 PM    No

If someone tries to lock tokens at 02:10 PM, since current stage at that point would be Stage 3, which isn't the beginning of epoch, we would be checking the allowed lock limits against the Stage 2 (since Stage 3 belongs to the same epoch), while the Stage 2 is effectively in future. There are probably many other messy situations similar to this one.

I would also expect that, in practice, we will most likely have at most one stage scheduled for the future timestamp.

dusan-maksimovic:
> Linking the Hydro contract with the gatekeeper contract can only happen during instantiation afaict. This probably needs to also allow setting/removing the gatekeeper by a whitelist admin afterwards, otherwise we'd have to migrate the contract to change the gatekeeper

I was thinking that the migration would be done by crating a new release with a migrate() function that just sends the Wasm msg to instantiate the Gatekeeper, and the reply logic is already there. True that the new version of SCs must be released in that case. Doesn't look very likely we are going to remove the Gatekeeper after it gets introduced, but I am also fine with have the add/remove mechanism.

dusan-maksimovic:
> Doesn't seem like the interchain tests are/were working on this one, trying to investigate that

They actually did, I've run them couple of times successfully. Not sure why they started failing...

PR #247, opened 2025-04-30
Add Vortex contract

p-offtermatt:
## Description

Closes: *#XXXX*

The purpose of Vortex is to allow Hydro to create concentrated liquidity positions of ATOM vs potentially volatile tokens while minimizing risk to Hydro’s ATOM principal. It does this by creating concentrated liquidity positions where the other side of the position acts as collateral to secure the ATOM side.
More description on: [vortex liquidity collateral system](https://www.notion.so/informalsystems/Vortex-Liquidity-Collateral-System-1c1d212b18c380ddbf54f904829fcf4d)

---

### Author Checklist

I have...

* [ ] Targeted the correct branch
* [ ] Included the necessary unit tests


contracts/liquid-collateral/documentation.md
```diff
@@ -0,0 +1,85 @@
+# Contract storage
+
+## State
+- project_owner: address which is optionally set during initialization. If set, only this address will have permission rights to execute 'create_position' msg.
```
p-offtermatt, line 4: why not just have these in the code itself? I am not sure this part of the doc does much

contracts/liquid-collateral/src/state.rs
```diff
@@ -0,0 +1,88 @@
...
+- round_end_time: defined during contract initialization - when the time ends - 'end_round' can be executed
+- auction_duration: duration of the potential auction set on initialization of the contract. If contract reaches the stage where auction is needed, this parameter will be utilized
+- auction_end_time: similarly,if auction is needed, auction_duration will be utilized to define auction_end_time
+- auction_principal_deposited: tracks the amount of deposited principal amount during bidding
+- principal_to_replenish: on the position creation, this parameter has the same value as initial_principal_amount, with the difference that this parameter is decremeneted as principal amount is being replenished
+- counterparty_to_give: determined if auction is needed - this is the amount of counterparty tokens contract can spend to fulfill the auction
+- position_rewards: helper field needed for storing the information about how much rewards contract will receive in case of liquidation - so that those funds can be distributed immediatelly to principal_funds_owner
+*/
```
p-offtermatt, line 27: Why not just put this as comments on the actual parts of the struct?

i.e. 

// If set, only this address has permission rights to execute `create_position` messages.
// Optionally set during instantiation
pub project_owner: ...

p-offtermatt (commented):
Just everywhere in general, docstrings have to be on the structs that they actually belong to, not just a block of text nearby

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+## End round bid
+ - can only be executed if auction is in progress
+ - bidder sends desired principal amount to the contract and request some amount of counterparty token
+ - bidder must replenish at least 1% of principal
+ - bidder cannot request more counterparty than contract has available
+ - bid is being saved in bids and in sorted bids
+ - in case the new bid makes principal amount be potentially replenished, unneeded (worst) bid/s will be kicked out
+ - if one or more bids are kicked out from sorted bids - bidders will be refunded and correct status of the bid will be saved
+*/
+pub fn end_round_bid(
```
p-offtermatt, line 685: This name isn't great imo, because it's a noun. all other execute messages are verbs. I would call it
place_bid or add_bid or similar

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    msg: EndRoundBidMsg,
+) -> Result<Response, ContractError> {
+    // Reject if user already has an existing bid
+    if BIDS.may_load(deps.storage, info.sender.clone())?.is_some() {
+        return Err(ContractError::BidAlreadyExists {});
+    }
+
```
p-offtermatt, line 695: why not allow an address to place more than one bid? it seems like a not-outlandish scenario to have someone place many bids to emulate some kind of "curve", i.e. a handful of progressively worse bids

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+        QueryMsg::Bids {} => to_json_binary(&query_get_bids(deps)?),
+        QueryMsg::SortedBids {} => to_json_binary(&query_sorted_bids(deps)?),
+    }
+}
+/*
+
+## Create position with reply
+ - based on the passed position arguments, method is sending MsgCreatePosition to the cl module on Osmosis
+ - in the reply contract is updating the state with the position information.
+*/
```
p-offtermatt, line 90:
```suggestion
/// Create position with reply
/// - based on the passed position arguments, method is sending MsgCreatePosition to the cl module on Osmosis
/// - in the reply contract is updating the state with the position information.
```

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+                token_min_amount0 = msg.counterparty_token_min_amount.to_string();
+            }
+        } else {
+            return Err(ContractError::AssetNotFound {});
+        }
+
+        tokens_provided.push(osmo_coin);
+    }
+
+    // Create position message
```
p-offtermatt, line 148: this is the type of comment that doesn't make sense (I guess it's just a remainder of sketching the code out) since it just says what the line after it states very plainly, so it doesn't add any value

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+
+    // Only owner can create position - if present
+    match &state.project_owner {
+        Some(owner) if info.sender != *owner => {
+            return Err(ContractError::Unauthorized {});
+        }
+        _ => {}
+    }
+
+    /*
```
p-offtermatt, line 112: this is a very useful comment for example, since it states something which otherwise is not obvious

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+    // Assign based on whether principal is first
+    if state.principal_first {
+        state.initial_principal_amount = amount0;
+        state.principal_to_replenish = amount0;
+        state.initial_counterparty_amount = amount1;
+    } else {
+        state.initial_principal_amount = amount1;
+        state.principal_to_replenish = amount1;
+        state.initial_counterparty_amount = amount0;
+    }
```
p-offtermatt, line 395: What happens to the remaining tokens? i.e. the provided tokens - the amounts that come back from the submsg?

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+     - update counteparty amount available in the auction
+     - start the auction
+*/
+pub fn end_round(deps: DepsMut, env: Env, _info: MessageInfo) -> Result<Response, ContractError> {
+    // Load the current state
+    let mut state = STATE.load(deps.storage)?;
+
+    let current_time = env.block.time;
+
+    // Check that the round is ended by checking that the round_id is less than the current round
```
p-offtermatt, line 629: outdated comment?

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+    let mut state = STATE.load(deps.storage)?;
+
+    let liquidity = state
+        .position_id
+        .map(|id| {
+            ConcentratedliquidityQuerier::new(&deps.querier)
+                .position_by_id(id)
+                .ok()
+                .and_then(|resp| resp.position) // FullPositionBreakdown
+                .and_then(|breakdown| breakdown.position) // Option<Position>
```
p-offtermatt, line 425: why are you writing out the type here? was that for debugging? not sure it serves a purpose, any IDE will let you inspect the types anyways

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+    } else {
+        // If principal is amount1, counterparty is amount0
+        Uint128::from_str(&response.amount0)
+    }
+    .map_err(|_| StdError::generic_err("Invalid Uint128 value"))?;
+
+    let liquidator_address = state
+        .liquidator_address
+        .clone()
+        .ok_or(ContractError::NoLiquidatorAddress {})?;
```
p-offtermatt, line 444: this errors if there is no liquidator? isn't there an error here because there are legitimate cases where there is no liquidator, but funds should still be sent?

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+
+    // Handle rewards if present
+    if let Some(rewards) = &state.position_rewards {
+        for reward in rewards {
+            // Query contract's balance for the denom
+            let contract_balance = deps
+                .querier
+                .query_balance(env.contract.address.clone(), &reward.denom)?;
+
+            // If contract has enough balance, create a send msg
```
p-offtermatt, line 467: when would the other case ever happen?

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+     - all fetched counterparty amount is sent to position_created_address (or project owner)
+     - potential excessive principal amount is sent to position_created_address
+     - exact amount needed to be replenished is sent to principal_funds_owner
+   - in case there were not enough principal amount for replenish:
+     - send whatever principal amount is fetched to principal_funds_owner
+     - decrement principal amount needed
+     - update counteparty amount available in the auction
+     - start the auction
+*/
+pub fn end_round(deps: DepsMut, env: Env, _info: MessageInfo) -> Result<Response, ContractError> {
```
p-offtermatt, line 623: it feels problematic that end_round could be called multiple times, right? is there any check for that?

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+            }
+        } else if coin.denom == state.counterparty_denom {
+            if state.principal_first {
+                token_min_amount1 = msg.counterparty_token_min_amount.to_string();
+            } else {
+                token_min_amount0 = msg.counterparty_token_min_amount.to_string();
+            }
+        } else {
+            return Err(ContractError::AssetNotFound {});
+        }
```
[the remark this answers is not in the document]
p-offtermatt: just to clarify, there are 2 concerns: checking the coin denoms (this should be here), and filling the min amounts (this should not be here)

contracts/liquid-collateral/src/state.rs
```diff
@@ -0,0 +1,88 @@
...
+
+/*
+## Bids
+ - official bids stored when bidder manages to create a bid in the auction period
+ - Bid fields: bidder (address), principal_deposited, tokens_requested (counterparty), tokens_fulfilled (counterparty), tokens_refunded(principal) and bid status.
+ - Bid status can be: Submitted, Processed and Refunded.
+ - since the bids can be partially fulfilled - the contract saves how much tokens are either fully/partially fulfilled, or refunded in case of bid not being one of the "winner" bids.
+ - The bid is initally Submitted, but in case bid is not needed afterwards - it will be refunded and the status will be set as Refunded. In case bid is considered as one of the 'winners' bid , it will be processed and the status will be Processed - but note that contract may not need the whole bid.
+
+## Sorted bids
```
p-offtermatt, line 64: this is missing by what the bids are sorted, and whether it's ascending or descending. that seems like a good invariant to note

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+
+            if collecting_bids {
+                accumulated_principal += *principal_deposited;
+                selected_bids.push(bid.clone());
+
+                if accumulated_principal >= state.principal_to_replenish {
+                    collecting_bids = false; // Stop collecting
+                }
+            } else {
+                if bidder == info.sender {
```
p-offtermatt, line 783: if we accomodate multiple bids per bidder, this needs to change, too

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+        counterparty_to_give: state.counterparty_to_give,
+        auction_principal_deposited: state.auction_principal_deposited,
+        position_rewards: state.position_rewards,
+        round_end_time: state.round_end_time,
+    };
+
+    Ok(response)
+}
+
+pub fn query_get_bids(deps: Deps) -> StdResult<Vec<(String, Bid)>> {
```
p-offtermatt, line 1066: this probably needs pagination

p-offtermatt: i.e. start_from and limit

contracts/liquid-collateral/src/contract.rs
```diff
@@ -0,0 +1,1089 @@
...
+        .add_attribute("pool_id", state.pool_id.to_string())
+        .add_attribute("lower_tick", msg.lower_tick.to_string())
+        .add_attribute("upper_tick", msg.upper_tick.to_string()))
+}
+
+/*
+## Liquidate position with reply
+ - method checks whether principal amount in the position is zero - which needs to be the case in order to allow liquidation
+ - the percentage of liquidation amount is calculated based on the principal funds liquidator sent to the contract (can be full or partial)
+ - amount is immediatelly transferred to the principal_funds owner (so contract doesn't need to hold anything)
```
p-offtermatt, line 179:
```suggestion
 - amount is immediately transferred to the principal_funds owner (so contract doesn't need to hold anything)
```

p-offtermatt (commented):
Nice work. I left some comments. I haven't looked into tests yet, that'll be up next.

contracts/liquid-collateral/src/state.rs
markoj2207: Resolved with [c3a4ff3377b4ab8afaaa803370c416edffac88a5](https://github.com/informalsystems/hydro/commit/c3a4ff3377b4ab8afaaa803370c416edffac88a5)

contracts/liquid-collateral/src/state.rs
```diff
@@ -0,0 +1,88 @@
...
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct State {
+    pub project_owner: Option<Addr>,
+    pub principal_funds_owner: Addr,
+    pub pool_id: u64,
+    pub position_created_address: Option<Addr>,
+    pub position_id: Option<u64>,
+    pub principal_denom: String,
+    pub principal_first: bool, // Flag to track in which position is the principal
```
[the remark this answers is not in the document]
markoj2207: Resolved with [c3a4ff3377b4ab8afaaa803370c416edffac88a5](https://github.com/informalsystems/hydro/commit/c3a4ff3377b4ab8afaaa803370c416edffac88a5)

contracts/liquid-collateral/src/contract.rs
markoj2207: Resolved with [c3a4ff3377b4ab8afaaa803370c416edffac88a5](https://github.com/informalsystems/hydro/commit/c3a4ff3377b4ab8afaaa803370c416edffac88a5)

contracts/liquid-collateral/src/state.rs
markoj2207: Resolved with [c3a4ff3377b4ab8afaaa803370c416edffac88a5](https://github.com/informalsystems/hydro/commit/c3a4ff3377b4ab8afaaa803370c416edffac88a5)

contracts/liquid-collateral/src/contract.rs
markoj2207: Resolved with [c3a4ff3377b4ab8afaaa803370c416edffac88a5](https://github.com/informalsystems/hydro/commit/c3a4ff3377b4ab8afaaa803370c416edffac88a5)

contracts/liquid-collateral/src/contract.rs
markoj2207: The idea with saving the liquidator address during "liquidate" action such that in the reply the corresponding funds can be distributed to liquidator.

With the current design - it should not be possible that liquidator address is not present which would mean something serious is not right.

But just in case, we have this check so that if this somehow happens - everything should be reverted.

contracts/liquid-collateral/src/contract.rs
markoj2207: If the end_round is sucessfully executed - that means the position is deleted on the Osmosis side.
If anyone tries to execute the end_round message again, it will fail since there is no position anymore.
In case it's preferable for us to have this check on our side - we can add it and this will only make the tx fail earlier.

contracts/liquid-collateral/src/contract.rs
markoj2207: This is now enabled in [486f34b69f625359618c0d461504f57f03e4046a](https://github.com/informalsystems/hydro/pull/247/commits/486f34b69f625359618c0d461504f57f03e4046a)

Every bid now has an id - so that one user can submit multiple bids.

contracts/liquid-collateral/src/contract.rs
markoj2207: The idea behind checking this was the concern that the contract might receive a bit less of reward token/s than stated in claimable rewards.

The reason for the concern is that when withdrawing from position (for example there was 10 principal and 15 counterparty amounts in the position) - the contract in most cases actually receive 9 and 14 (one less - probably because of rounding).

For rewards it was never the case - it was always the same amount as stated. 
So the check is removed in https://github.com/informalsystems/hydro/pull/247/commits/486f34b69f625359618c0d461504f57f03e4046a

contracts/liquid-collateral/src/contract.rs
markoj2207: Yes, changed in https://github.com/informalsystems/hydro/pull/247/commits/486f34b69f625359618c0d461504f57f03e4046a

contracts/liquid-collateral/src/contract.rs
markoj2207: Added in https://github.com/informalsystems/hydro/pull/247/commits/486f34b69f625359618c0d461504f57f03e4046a

contracts/liquid-collateral/src/contract.rs
p-offtermatt: Yep, that makes sense!

contracts/liquid-collateral/src/contract.rs
p-offtermatt: Ok, yep, that makes sense. You route the reply based on id, that seems good

contracts/liquid-collateral/src/contract.rs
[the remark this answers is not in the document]
markoj2207: This comes from the fact we need know the exact place of tokens in the cl pool.

The loop basically goes through provided  tokens and forces that the denom must be either principal or counterparty. Since no duplicate denoms are allowed by cosmwasm - this will mean there can either be one or two denoms inside provided funds.

For the min amounts:
Since inside contract's create position msg user specifies principal min amount and counterparty min amount - we need to know how to send them inside osmosis' create position msg. (token_min_amount0 and token_min_amount1) - in terms which one goes first (as 0 place).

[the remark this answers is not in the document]
markoj2207: Resolved with [9ca2d1e8f24255740a0c9d2313ca25f7d8425542](https://github.com/informalsystems/hydro/pull/247/commits/9ca2d1e8f24255740a0c9d2313ca25f7d8425542)

contracts/liquid-collateral/src/contract.rs
markoj2207: Resolved with [b24440c82068ed85f7e9d50b00d882467b246b7a](https://github.com/informalsystems/hydro/pull/247/commits/b24440c82068ed85f7e9d50b00d882467b246b7a)

contracts/liquid-collateral/src/state.rs
```diff
@@ -0,0 +1,88 @@
+use cosmwasm_std::{Addr, Coin, Decimal, Timestamp, Uint128};
+use cw_storage_plus::{Item, Map};
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+#[derive(Serialize, Deserialize, Debug)]
+pub struct State {
+    // Address which is optionally set during initialization. If set, only this address will have permission rights to execute 'create_position' msg.
+    pub project_owner: Option<Addr>,
```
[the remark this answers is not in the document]
markoj2207: Resolved with: [https://github.com/informalsystems/hydro/pull/247/commits/ecdb6dc1ce54ffde45aaed99305f4c0a24296c7e](url)

The user which instantiates the contract can now optionally specify: position_admin and counterparty_owner.

If both parameters are not set during init, counterparty_owner will become the address which made contract enter the position.

PR #252, opened 2025-05-08
Fix tribute migration and UserVotes query

arlai-mk:
## Description

Closes: *#XXXX*

This is a continuation of `moonkitt/nft-preparation` work that was already merged into Hydro `main` branch, after we realized that we don't need to query the votes per lock ID, as we just need the original voters to claim the rewards for what they voted themselves. and the rewards don't follow the lock IDs.

What this update does:
1. It reverts tribute contract migration code so the tribute contract is mostly unchanged (just cleaning an old migration artifacts)
2. It takes into account that the USER_LOCKS snapshots did not start from the beginning of Hydro being live on mainnet. `query_user_votes` will use `VOTE_MAP_V1` for any round before the snapshots were added, and `VOTE_MAP_V2` afterwards, so that all users are still able to claim.

---

### Author Checklist

I have...


contracts/tribute/src/contract.rs
```diff
@@ -167,92 +174,40 @@ fn claim_tribute(
...
     // make sure that tributes for this proposal are claimable
-    get_proposal_tributes_info(
-        &deps.as_ref(),
-        &config,
-        round_id,
-        tranche_id,
-        tribute.proposal_id,
-    )?
-    .are_tributes_claimable()?;
+    get_proposal_tributes_info(&deps.as_ref(), &config, round_id, tranche_id, vote.prop_id)?
```
dusan-maksimovic, line 198: Let's use `tribute.proposal_id` here instead of `vote.prop_id`.

contracts/tribute/src/contract.rs
```diff
@@ -167,92 +174,40 @@ fn claim_tribute(
...
+        .are_tributes_claimable()?;
 
-    let proposal = get_proposal(
-        &deps.as_ref(),
-        &config,
-        round_id,
-        tranche_id,
-        tribute.proposal_id,
-    )?;
+    let proposal = get_proposal(&deps.as_ref(), &config, round_id, tranche_id, vote.prop_id)?;
```
dusan-maksimovic, line 201: Same `tribute.proposal_id` here.

contracts/tribute/src/contract.rs
```diff
@@ -533,23 +492,25 @@ fn query_proposal(
...
             round_id,
             tranche_id,
             address,
         },
     )?;
 
-    Ok(user_voted_locks)
+    let vote = user_vote_resp.votes;
+
+    Ok(vote)
```
dusan-maksimovic, line 513:
```suggestion
    Ok(user_vote_resp.votes)
```

dusan-maksimovic (commented):
LGTM, just minor comments.

contracts/tribute/src/contract.rs
arlai-mk: Done

contracts/tribute/src/contract.rs
arlai-mk: Done

contracts/tribute/src/contract.rs
arlai-mk: Done

dusan-maksimovic (approved):

PR #259, opened 2025-06-04
stTOKEN ratio update implementation

dusan-maksimovic:
## Description

Closes: *#255*

This PR adds support for periodically fetching the stTOKEN ratio from Stride chain and updating its value in Hydro contracts. Neutron Interchain Queries are leveraged to achieve this goal, which means that the current implementation of the ST Token Info Provider is tightly coupled to Neutron chain. Interchain Query used in this code queries the state of the "stakeibc" module to obtain the Host Zone information. Host zones (on Stride chain) represent cosmos chains for whose native tokens the stTOKENs are issued through Stride protocol (e.g. Cosmos Hub, Celestia, Dymension, etc.). Host zone contains information about the redemption rate of the stTOKEN against the underlying token.

Two new execute() variants are added to ST Token Info Provider SC: one to register ICQ query and the other one to remove it. First one is permission-less, while the second one can only be executed by the wallet that created the ICQ. This way a single ICQ existence is enforced, and also helps with testing. The `InterchainQueryInfo` query is only used by our internal tool that is responsible for registering the ICQ and relaying the results.

---

### Author Checklist

I have...


p-offtermatt (approved):
LGTM, thanks for the work

PR #281, opened 2025-07-15
Lock tracking

markoj2207:
## Description

This update introduces internal tracking for lock operations and adds depth validation to prevent excessively nested lock structures.

### Changes:
- **State additions:**
  - `LOCK_ID_TRACKING`: tracks forward composition (split/merge relationships).
  - `REVERSE_LOCK_ID_TRACKING`: maps each lock to its parent(s).
  - `LOCK_ID_EXPIRY`: stores expiry timestamps for lock IDs. (stores the timestamp of lock deletion from storage)

- **Constants added:**
  - `LOCK_EXPIRY_DURATION_SECONDS`: defines how long a lock remains valid.
  - `LOCK_DEPTH_LIMIT`: sets a maximum allowed ancestor depth for any lock.
**NOTE - we should revisit current constant values and update it to the desired values if necessary**

- **Execution updates:**
  - `SplitLock` and `MergeLocks` now populate the tracking maps and enforce depth checks based on ancestry.

- **Utility functions:**
  - `get_current_lock_composition`: recursively resolves the composition of a lock into its leaf components.
  - `get_lock_ancestor_depth`: calculates the depth of non-expired ancestors for a given lock.

### Author Checklist

I have...


contracts/hydro/src/contract.rs
```diff
@@ -90,6 +91,10 @@ pub const MIN_DEPLOYMENT_DURATION: u64 = 1;
 
 pub const MIN_SPLIT_LOCK_SIZE: Uint128 = Uint128::new(10_000);
 
+pub const LOCK_EXPIRY_DURATION_SECONDS: u64 = 60 * 60 * 24 * 30; // 30 days
```
dusan-maksimovic, line 94: I think both of these constants should be moved into `Constants` so they can be updated without need to store the new codes and perform migration.

Also the `update_config()` should be modified to support update of those values.

Keep in mind that this change also requires migration of Constants in our deployed contracts.

contracts/hydro/src/contract.rs
```diff
@@ -90,6 +91,10 @@ pub const MIN_DEPLOYMENT_DURATION: u64 = 1;
...
+
+pub const LOCK_DEPTH_LIMIT: u32 = 5;
```
p-offtermatt, line 96: make this 50 in the migration

contracts/hydro/src/contract.rs
p-offtermatt, line 94: let's make this 6 months

contracts/hydro/src/contract.rs
```diff
@@ -989,9 +1027,32 @@ fn merge_locks(
         env.block.height,
     )?;
 
+    let mut parents = REVERSE_LOCK_ID_TRACKING
```
p-offtermatt, line 1030: this should always be empty, no need to load from store?

PR #293, opened 2025-07-24
get_lock_ancestor_depth optimization

markoj2207:
## Description

This PR introduces an in-memory cache to the get_lock_ancestor_depth function to avoid redundant computations when traversing shared ancestor paths.

---

### Author Checklist

I have...


p-offtermatt (commented):
One obvious optimization is to have get_lock_ancestor_depth return the depth map (essentially the cache), then use that cache to initialize the calls for the other locks in the merge_locks call.

To get the depth of a node, you simply query the returned map for that node, and its guaranteed to have a value in the cache if you passed it into the call as the lock_id

markoj2207:
> One obvious optimization is to have get_lock_ancestor_depth return the depth map (essentially the cache), then use that cache to initialize the calls for the other locks in the merge_locks call.
> 
> To get the depth of a node, you simply query the returned map for that node, and its guaranteed to have a value in the cache if you passed it into the call as the lock_id

Yes, actually wanted to suggest this after we see how will optimization inside get_lock_ancestor_depth work - but we can actually have this in this go. 
This should make us having only one "heavy gas usage" call to the method - no matter how many times it was called from the loop.

PR #307, opened 2025-09-01
Inflow withdrawal workflow implementation

dusan-maksimovic:
[Recompile the contracts and regenerate the schema before merging]

## Description

This PR adds 4 new execute() variants, as well as several new queries to the Inflow contract, in order to implement the funds withdrawal functionality. New actions:
- `Withraw {}` - Users initiate withdrawals with this action. If there are enough funds to cover all pending withdrawals plus the new one, users will receive the deposit tokens immediately. Otherwise, they enter the withdrawal queue.
- `CancelWithrawal { withdrawal_ids: Vec<u64> }` - Users can cancel their existing withdrawal request that are still in the queue, until the moment the funds for the given request have been provided.
- `FulfillPendingWithdrawals { limit: u64 }` - Permissionless action. Iterates over withdrawal requests queue, checks the smart contract balance, and marks the requests as funded if there is enough funds in the contract to pay them out to the users.
- `ClaimUnbondedWithdrawals { withdrawal_ids: Vec<u64> }` - Permissionless action. Iterates over the provided withdrawal request ids, checks if they have been funded and performs the actual payouts to those users that initiated those withdrawals.

---

### Author Checklist

I have...

### DIFF
--- a/contracts/inflow/src/contract.rs
+++ b/contracts/inflow/src/contract.rs
@@ -1,12 +1,16 @@
-use std::env;
+use std::{
+    collections::{HashMap, HashSet},
+    env, vec,
+};
 
 use cosmos_sdk_proto::cosmos::bank::v1beta1::{DenomUnit, Metadata};
 use cosmwasm_std::{
     entry_point, from_json, to_json_binary, to_json_vec, Addr, AnyMsg, BankMsg, Binary, Coin,
-    CosmosMsg, Decimal, Deps, DepsMut, Env, MessageInfo, Order, Reply, Response, StdError,
-    StdResult, SubMsg, Uint128,
+    ConversionOverflowError, CosmosMsg, Decimal, Deps, DepsMut, Env, Int128, MessageInfo, Order,
+    Reply, Response, StdError, StdResult, Storage, SubMsg, Uint128,
 };
 use cw2::set_contract_version;
+use cw_storage_plus::Bound;
 use neutron_sdk::{
     bindings::{msg::NeutronMsg, query::NeutronQuery},
     proto_types::osmosis::tokenfactory::v1beta1::MsgSetDenomMetadata,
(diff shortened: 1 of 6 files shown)


contracts/inflow/src/contract.rs
```diff
@@ -100,6 +129,16 @@ pub fn execute(
 
     match msg {
         ExecuteMsg::Deposit {} => deposit(deps, env, info, &config),
+        ExecuteMsg::Withraw {} => withdraw(deps, env, info, &config),
+        ExecuteMsg::CancelWithrawal { withdrawal_ids } => {
```
markoj2207, line 133: Typos - should be: Withdraw and CancelWithdrawal.

contracts/inflow/src/contract.rs
```diff
@@ -139,6 +178,443 @@ fn deposit(
...
+    let mut shares_burned = Uint128::zero();
+    let mut amount_to_withdraw = Uint128::zero();
+
+    for withdrawal_id in &withdrawals_to_process {
+        // USER_WITHDRAWAL_REQUESTS should always be in sync with WITHDRAWAL_REQUESTS.
+        // If the withdrawal entry is not found, the entire execute() action should fail.
+        let withdrawal_entry = WITHDRAWAL_REQUESTS.load(deps.storage, *withdrawal_id)?;
+
+        // Double check that the withdrawal belongs to the sender and has not been funded yet
+        if withdrawal_entry.withdrawer != info.sender || withdrawal_entry.is_funded {
```
markoj2207, line 350: Do we really need to have a separate is_funded field to check - since there is already a LAST_FUNDED_WITHDRAWAL_ID storage used for filtering out the withdrawals which are already funded?

dusan-maksimovic: For the SC itself- we don't. But it is more explicit this way. For example, `query_user_withdrawal_requests()` returns all user requests, and this flag on each `WithdrawalEntry` comes in handy to have this info on the FE.

contracts/inflow/src/contract.rs
```diff
@@ -486,3 +1171,151 @@ fn create_set_denom_metadata_msg(
...
 }
+
+/// Updates the list of user's pending withdrawal requests by adding and/or removing specified withdrawal IDs.
+fn update_user_withdrawal_requests_info(
+    storage: &mut dyn Storage,
+    user: Addr,
+    config: &Config,
+    withdrawals_to_add: Option<Vec<u64>>,
+    withdrawals_to_remove: Option<Vec<u64>>,
+    should_check_withdrawal_limit: bool,
```
dusan-maksimovic, line 1182: I added this check to be optional for the following reason: If user already has maximum allowed withdrawal requests (let's say 10) and we `update_config()` to 5 afterwards, we wouldn't be able to execute `cancel_withdrawal()` and `claim_unbonded_withdrawals()` for this user, if we try to remove less than 5 of its withdrawal requests. It is a bit far fetched, but better to be safe than sorry.

contracts/inflow/src/contract.rs
```diff
@@ -288,14 +288,25 @@ fn withdraw(
...
 fn cancel_withdrawal(
     deps: DepsMut<NeutronQuery>,
+    env: Env,
     info: MessageInfo,
     config: &Config,
     withdrawal_ids: Vec<u64>,
 ) -> Result<Response<NeutronMsg>, ContractError> {
+    let total_pool_value =
+        get_total_pool_value(&deps.as_ref(), &env, config.deposit_denom.clone())?;
+    if total_pool_value == config.deposit_cap {
```
p-offtermatt, line 304: checking for equality seems problematic/wrong here, no?

PR #330, opened 2025-11-05
Multi-asset Inflow vaults implementation

dusan-maksimovic:
## Description

This PR adds support for depositing multiple assets into a single Inflow vault. Main changes:
- Introduction of a new smart contract called Control Center that can manage multiple Inflow smart contract instances. Having multiple Inflow SC instances allows us to accept different deposit tokens, while having them bundled with the Control Center allows us to control how many shares each deposit is worth.
- Inflow contract now issues vault shares based on the total vault value, total number of shares issued by all Inflow instances managed by common Control Center, and also by taking into account the ratio of the deposit token towards the base token (e.g. ATOM).
- Inflow contract is also connected to the token info provider contract, which allows us to obtain the ratio of the deposit token to the base token. Since we also support base token deposits, if a given Inflow instance doesn't have a reference to the token info provider contract, it is assumed that it accepts deposits of the base token.
- `DEPLOYED_AMOUNT` store is moved to the Control Center, and it is maintained in base token units (e.g. ATOMs).
- Migration is added in order to be able to migrate existing Inflow instances to new setup. We will first instantiate the Control Center by providing addresses of existing vaults, and then pass its address through the Inflow MigrateMsg.
- Even if our vault will still accept only one deposit denom (e.g. USDC) we will still have the setup that includes the Control Center, to keep common code base and also to be able to introduce other USDx variants in the future.  
- Existing dTOKEN and stTOKEN information provider contracts have been modified to expose additional query that will be used by the Inflow contract to query the ratio to the base token. This way we can reuse existing TIP instance that are having the token ratios updated for the purpose of Hydro.
- `Inflow` SC UTs have been re-worked to use dATOM as deposit token, in order to catch any ratio-related issues more easily.

---

### Author Checklist

I have...


p-offtermatt:
What happens if a sub vault gets removed? That subvault contract becomes very messed up, right?
i.e. it would mint/burn shares at weird ratios, since it still queries the control center.

I'm wondering what that means for removing/adding (adding makes more sense); I guess it would only make sense if the subvault is also migrated to new code at the same time as its being removed

PR #363, opened 2026-01-04
Add query for control center pool info right from the vault contract

p-offtermatt:
This skips a step when just trying to query the ratio for a single pools shares.


p-offtermatt:
@copilot add a unit test for the new query

p-offtermatt:
Clippy seems unrelated though, hm.

@copilot create a new issue to fix the Clippy problems surfaced from this PR, then assign that issue to yourself and create a PR

PR #398, opened 2026-03-08
Fix LSM denom resolution broken by Neutron v10 / ibc-go v10 upgrade

### LINKED ISSUE #397: LSM denom resolution broken after Neutron v10 upgrade (ibc-go v10 removed DenomTrace gRPC method)
## Summary

All LSM tokenized share lockups currently report zero voting power and `can_lock_denom` returns `false` for any LSM denom. This is caused by the Neutron v10.0.0 mainnet upgrade (height 50040000, Feb 25 2026) which moved to ibc-go v10.5.0 — removing the `DenomTrace` gRPC endpoint that Hydro relies on.

## Root Cause

`lsm_integration.rs` hardcodes the gRPC path:
```rust
pub const DENOM_TRACE_GRPC: &str = "/ibc.applications.transfer.v1.Query/DenomTrace";
```
This endpoint existed through ibc-go v8 but was removed in ibc-go v9+. The replacement is the `Denom` method (`/ibc.applications.transfer.v1.Query/Denom`), which takes `{"hash": "ibc/..."}` and returns the trace under a different response structure.

## Evidence

- `can_lock_denom` returns `false` for e.g. `ibc/02C8EE465C8A552CB20E537394F69216A9FE5C427BD4B68E4FFD1ABB56DE53F8` (Informal Systems LSM share, cosmosvaloper16k.../92743 via transfer/channel-1) — a denom that was successfully locked and voted in rounds 6, 10, and 13.
- `grpcurl` against Neutron mainnet confirms `DenomTrace` is absent; only `Denom`, `Denoms`, `DenomHash`, `TotalEscrowForDenom` are available.
- Neutron v9.0.0 (Nov 10 2025) used ibc-go v8.7.0 (DenomTrace present). Neutron v10.0.0 (Feb 25 2026) uses ibc-go v10.5.0 (DenomTrace gone).

## Impact

Every LSM lockup's voting power computation via `lockup_voting_metrics` errors out with the generic "Token with denom ... can not be locked in Hydro." message. `validate_denom` fails for all LSM denoms, so `get_token_denom_ratio` returns zero for all of them, meaning all LSM-based lockups have zero voting power in round 16+.

## Fix

Update `query_ibc_denom_trace` in `lsm_integration.rs` to use the new `Denom` endpoint and its corresponding request/response proto types.

p-offtermatt:
Closes #397

## Summary

- Replaces the removed `DenomTrace` gRPC endpoint (`/ibc.applications.transfer.v1.Query/DenomTrace`) with the new `Denom` endpoint (`/ibc.applications.transfer.v1.Query/Denom`) introduced in ibc-go v10
- Defines the new proto request/response types inline (not yet in the `ibc-proto` crate)
- Converts the new `Denom` response back to the legacy `DenomTrace` format so all downstream validation logic in `resolve_validator_from_denom` is unchanged
- Updates the test mock (`testing_mocks.rs`) to encode/decode the new proto format

## Context

Neutron v10.0.0 (mainnet height 50040000, Feb 25 2026) upgraded to ibc-go v10.5.0, which removed the `DenomTrace` gRPC method. This broke all LSM denom resolution: `validate_denom` fails for every LSM share, causing `can_lock_denom` to return `false` and all LSM lockups to report zero voting power in round 16+.

The new `Denom` endpoint lives in the same `ibc.applications.transfer.v1.Query` service and accepts the same hash input. The `Denom` response uses a structured `trace: Vec<Hop>` instead of the old flat `path` string; the conversion is straightforward (`port_id/channel_id` joined with `/`).

## Test plan

- [x] `make clippy` passes
- [x] `make test-unit` passes (all existing LSM denom resolution tests cover the updated mock)
- [ ] `make compile` (requires Docker)

### DIFF
--- a/.changelog/unreleased/bug-fixes/397-fix-denom-trace-grpc.md
+++ b/.changelog/unreleased/bug-fixes/397-fix-denom-trace-grpc.md
@@ -1,3 +1,3 @@
 - Fix LSM denom resolution broken by Neutron v10 upgrade: replace the removed
   `DenomTrace` gRPC endpoint with the new `Denom` endpoint introduced in ibc-go v10.
-  ([\#397](https://github.com/informalsystems/hydro/pull/397))
+  ([\#398](https://github.com/informalsystems/hydro/pull/398))


contracts/hydro/src/lsm_integration.rs
```diff
@@ -1,17 +1,47 @@
...
+    #[prost(message, repeated, tag = "3")]
+    pub trace: Vec<DenomHop>,
+}
+
+#[derive(Clone, PartialEq, Message)]
+pub(crate) struct QueryDenomResponse {
+    #[prost(message, optional, tag = "1")]
+    pub denom: Option<IbcDenom>,
+}
+
```
dusan-maksimovic: We should not redefine these data structures. Instead, we should bump `ibc-proto` crate to 0.52.0, since this version contains new data structures that match the updated API.

dusan-maksimovic:
I agree that the fix should be to update the endpoint we are querying to resolve the denom. But the approach in this PR isn't the best. See my comment above.

PR #421, opened 2026-06-22
Inflow on Cosmos Hub

arlai-mk:
## Description

Support Inflow contracts on Cosmos Hub for upcoming migration from Neutron, and include a `Pause` / `Unpause` capability on the Vaults.

---

### Author Checklist

I have...


p-offtermatt:
**`AvailableForDeposit` behaves differently across the three adapters — intentional?**

`cctp-adapter-cosmoshub`'s `query_available_for_deposit` (contract.rs ~L883) now validates the depositor and denom, returning `Uint128::MAX` only when the depositor is registered + enabled and `denom == config.denom`, else `0`:

```rust
let amount = match depositor {
    Some(d) if d.enabled && denom == config.denom => Uint128::MAX,
    _ => Uint128::zero(),
};
```

That diverges from the two existing implementations, which ignore the params and always return `MAX`:
- Neutron `cctp-adapter` — `query_available_for_deposit()` returns `Uint128::MAX` (old comment: _"no cap for CCTP adapter"_).
- `basic-adapter-cosmoshub` — `AvailableForDeposit { .. } => MAX` unconditionally.

Two things I wanted to check:

1. **Was the new validating behavior intentional?** It actually looks _more_ correct to me — it's the only one of the three that honors the interface contract (`AdapterInterfaceQueryMsg::AvailableForDeposit` is documented as "the maximum amount that can be deposited [accounting for] caps and other limitations", and it takes `depositor_address`/`denom` precisely so an impl can use them). But the change is undocumented: the Neutron adapter's explanatory comment was dropped and nothing replaced it, so it reads as incidental rather than deliberate.

2. **Should the Neutron `cctp-adapter` and `basic-adapter-cosmoshub` be brought in line?** If `0`-for-unregistered/wrong-denom is the intended semantics, the inconsistency is a latent footgun — anything that routes deposits off this query will treat the three adapters differently.

Not blocking from my side — mainly want to confirm it's a conscious decision and, if so, add a short comment so the three adapters don't silently drift.

p-offtermatt (approved):
LGTM! Just one comment on some potential inconsistency between the adapters. Other than that, no complaints

contracts/inflow/shares-converter/src/msg.rs
```diff
@@ -0,0 +1,59 @@
...
+
+#[cw_serde]
+pub struct InstantiateMsg {
+    pub admin: String,
+    /// Optional initial conversion pairs to register at instantiation.
+    pub pairs: Vec<ConversionPair>,
+}
+
+#[cw_serde]
+pub enum ExecuteMsg {
```
dusan-maksimovic, line 17: Do we need a variant to extract remaining `cosmos_hub_shares_denom`s tokens after conversion is done? Could be useful in case there are users who never convert. Or we keep this SC live until Neutron is running and allow for indefinite conversion period?

arlai-mk:
> **`AvailableForDeposit` behaves differently across the three adapters — intentional?**
> ...
> 
> Two things I wanted to check:
> 
> 1. **Was the new validating behavior intentional?** It actually looks _more_ correct to me — it's the only one of the three that honors the interface contract (`AdapterInterfaceQueryMsg::AvailableForDeposit` is documented as "the maximum amount that can be deposited [accounting for] caps and other limitations", and it takes `depositor_address`/`denom` precisely so an impl can use them). But the change is undocumented: the Neutron adapter's explanatory comment was dropped and nothing replaced it, so it reads as incidental rather than deliberate.
> 2. **Should the Neutron `cctp-adapter` and `basic-adapter-cosmoshub` be brought in line?** If `0`-for-unregistered/wrong-denom is the intended semantics, the inconsistency is a latent footgun — anything that routes deposits off this query will treat the three adapters differently.
> 
> Not blocking from my side — mainly want to confirm it's a conscious decision and, if so, add a short comment so the three adapters don't silently drift.

I think the Neutron's CCTP adapter could need it too, I actually forgot to update it, but we could argue that we may not need to update as it's short-lived (it should be stopped in ~2 months).

For Basic adapter, we accept any denom, so we can't restrict on it. But we could also return 0 for a non-registered depositor.

contracts/inflow/shares-converter/src/msg.rs
arlai-mk: I think we should keep this SC live indefinitely, I don't really see any benefit of removing it (it doesn't cost anything to keep it live). Unless we want specifically to give users a specific timeline to convert, and if they don't convert during this time they lose their funds (and we increase the share ratio by burning the unclaimed shares)?

@p-offtermatt what would you think of it?

dusan-maksimovic (approved):
LGTM!

Except for one thing I posted before (but more as a question), everything looks very solid!

PR #425, opened 2026-07-10
Hydro migration to Cosmos Hub

dusan-maksimovic:
## Description

This PR introduces changes to SCs and adds off-chain tools needed to perform migration of Hydro to the Hub.
Main changes are:
1. Hydro contract is adjusted so it can work with new version of LSM token info provider that will be deployed on the Hub. The difference compared to the old one is that we no longer need to resolve IBC denoms of LSM tokens, since we will be on the Hub already.
2. Implemented LSM Hub and stATOM token info providers suitable for Hub deployment.
3. Extended Hydro SC with queries and executes needed to support the migration process.
4. Implemented off-chain tool that will be used to update validator power ratios on the new LSM Hub token info provider.
5. Implemented first part of a migration tool that reads the state of the current Hydro SC instance on Neutron and prepares data for further use during the migration process.

---

### Author Checklist

I have...


Cargo.toml
```diff
@@ -35,7 +37,7 @@ version = "3.6.12"
 edition = "2021"
 
 [profile.release]
-opt-level = 3
+opt-level = "s"
```
dusan-maksimovic, line 40: Instructs compiler to optimize for binary size, which reduces function in-lining, so we don't go over the limit.

tools/hydro-hub-migrations-tool/main.go
```diff
@@ -0,0 +1,316 @@
+package main
```
dusan-maksimovic, line 1: Tool that will read the state of existing SC, export the data, then use this data to migrate lockups to a new Hydro SC deployed on the Hub.

tools/hub-validators-tool/main.go
```diff
@@ -0,0 +1,394 @@
+package main
```
dusan-maksimovic, line 1: Similar to our ICQ tool except that it submits validator ratios to the Hub LSM provider contract.

contracts/hydro/src/ibc_transfer.rs
```diff
@@ -0,0 +1,90 @@
+use cosmwasm_std::{AnyMsg, Coin, CosmosMsg};
```
dusan-maksimovic, line 1: All the logic needed to send Neutron-specific CosmosMsg, but without adding neutron-std dependency, since we want the same code to be deployed on the Hub.

contracts/hydro/src/token_manager.rs
```diff
@@ -134,6 +134,8 @@ impl TokenManager {
 pub enum TokenInfoProvider {
     #[serde(rename = "lsm")]
     LSM(TokenInfoProviderLSM),
+    #[serde(rename = "lsm_hub")]
+    LSMHub(TokenInfoProviderLSMHub),
```
dusan-maksimovic, line 138: Different version for the Hub is needed since we don't need to resolve IBC denom for LSM tokens anymore. After migration is completed we should be able to remove the old `TokenInfoProviderLSM` completely.

contracts/hydro/src/token_manager.rs
```diff
@@ -367,6 +376,97 @@ impl TokenInfoProviderLSM {
...
+    // Returns OK if the denom is a valid LSM tokenized share on the Cosmos Hub
+    // of a validator that is also among the top max_validators validators
+    // for the given round, and returns the address of that validator.
+    pub fn resolve_denom(
+        &mut self,
+        deps: &Deps,
+        round_id: u64,
+        denom: String,
+    ) -> StdResult<String> {
+        let validator = extract_validator_from_lsm_denom(denom)?;
```
dusan-maksimovic, line 396: Here, we can extract the validator address directly from the denom, since we are on the Hub.

contracts/hydro/src/contract.rs
```diff
@@ -3096,6 +3210,10 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> Result<Binary, ContractErro
         QueryMsg::LockupsPendingSlashes { lockup_ids } => {
             to_json_binary(&query_lockups_pending_slashes(deps, lockup_ids)?)
         }
+        QueryMsg::AllLockups {
```
dusan-maksimovic, line 3219: Used by the migration tool.

contracts/token-info-providers/lsm-hub-token-info-provider/src/validators.rs
```diff
@@ -0,0 +1,401 @@
...
+use crate::{
+    error::ContractError,
+    msg::ExecuteContext,
+    state::{VALIDATORS_INFO, VALIDATORS_PER_ROUND},
+    utils::COSMOS_VALIDATOR_PREFIX,
+};
+
+const STAKING_VALIDATOR_GRPC: &str = "/cosmos.staking.v1beta1.Query/Validator";
+
+pub fn update_validators_ratios(
```
dusan-maksimovic, line 26: The logic here is more-less the same as it is in our existing LSM token info provider. The main difference is that now we can query the validator by its address directly and proceed with our logic immediately within the same TX.

contracts/token-info-providers/trusted-derivative-token-info-provider/src/contract.rs
```diff
@@ -0,0 +1,265 @@
+use cosmwasm_std::{
```
dusan-maksimovic, line 1: This one is for handling stATOM token ratio. As we agreed, a trusted address will periodically submit the ratio to this SC.

contracts/token-info-providers/trusted-derivative-token-info-provider/src/contract.rs
```diff
@@ -0,0 +1,265 @@
...
+                "failed to query current round id: {e}"
+            )))
+        })?;
+
+    initialize_token_ratio(deps.storage, current_round)?;
+
+    let old_ratio = TOKEN_RATIO.load(deps.storage, current_round)?;
+    let mut submsgs = vec![];
+
+    if old_ratio != new_ratio {
```
p-offtermatt, line 101: let's have a generous safeguard here that checks that the new ratio is not too far from the original ratio (e.g. let's see what 10 weeks of inflation would do to the ratio, then put that as the safeguard and expect that we run every week). if we need to update the ratio more than this (e.g. set it to 0 because stATOM became worthless, or we forgot to relay for more than 10 weeks), we can always do it via contract migration. This gives a bit of safety against a compromised trusted key

p-offtermatt (approved):
LGTM. One comment about the safeguard, but otherwise great. Just noting too that we need to run this on staging contracts first to rehearse everything, since there's a lot of moving pieces.

contracts/token-info-providers/trusted-derivative-token-info-provider/src/contract.rs
dusan-maksimovic: Resolved in https://github.com/informalsystems/hydro/pull/425/commits/e648511879a59027f6ef3aae57bd627e92f12d87

PR #426, opened 2026-07-24
Hydro contract adjustments for Hub deployment & ability to migrate lockups from Neutron

dusan-maksimovic:
## Description

Main changes:
1. Removed Neutron-specific LSM handling code from Hydro smart contract and switched to using Hub-designed variant
2. Removed messages and queries specific for Neutron that are no longer applicable (TransferFundsToHub and everything related to dATOM conversion)
3. Cleaned-up SC state to remove stores that aren't needed for the Hub deployment.
4. Implemented MintLockups() message that allows us to migrate existing lockups from Neutron instance
5. Implemented off-chain tool command that allows us to submit transactions to mint lockups on the Hub instance.

---

### Author Checklist

I have...


contracts/hydro/src/contract.rs
```diff
@@ -641,6 +612,140 @@ pub fn lock_tokens(
...
+// not go through the Gatekeeper.
+fn mint_lockups(
+    mut deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    constants: &Constants,
+    lockups_to_mint: Vec<LockupToMint>,
+) -> Result<Response, ContractError> {
+    // validate that the sender is on the whitelist
+    let whitelist = WHITELIST.load(deps.storage)?;
```
p-offtermatt, line 629: Sure that it's fine to let whitelist addresses do it? admins seems better

contracts/hydro/src/contract.rs
```diff
@@ -641,6 +612,140 @@ pub fn lock_tokens(
...
+
+    let current_round = compute_current_round_id(&env, constants)?;
+    let mut token_manager = TokenManager::new(&deps.as_ref());
+
+    let mut total_locked_tokens = LOCKED_TOKENS.load(deps.storage)?;
+    let mut seen_lock_ids: HashSet<u64> = HashSet::new();
+    let mut lock_ids_minted = vec![];
+
+    for lockup in lockups_to_mint {
+        if !seen_lock_ids.insert(lockup.lock_id) {
```
p-offtermatt, line 642: check that minted lock ids are below the LOCK_ID counter?

contracts/hydro/src/lsm_integration.rs
```diff
@@ -1,45 +1,12 @@
...
 
-use ibc_proto::ibc::apps::transfer::v1::{DenomTrace, QueryDenomRequest, QueryDenomResponse};
-
-use prost::Message;
-
-pub const IBC_TOKEN_PREFIX: &str = "ibc/";
 // ibc-go v10 (shipped in Neutron v10) removed the DenomTrace gRPC endpoint and replaced it
 // with a new Denom endpoint. The new endpoint is available at the following path.
 pub const DENOM_GRPC: &str = "/ibc.applications.transfer.v1.Query/Denom";
 pub const TRANSFER_PORT: &str = "transfer";
```
p-offtermatt: These two constants are now dead in production code: `query_ibc_denom_trace()` and `resolve_validator_from_denom()` were the only consumers, and both are removed in this PR. `DENOM_GRPC` is only referenced by `testing_mocks.rs`, which mocks a gRPC query the contract no longer issues, and `TRANSFER_PORT` has no remaining users in this crate. The ibc-go v10 / Neutron comment above is stale for the Hub deployment as well. Suggest deleting both constants (plus the corresponding mock plumbing in `testing_mocks.rs`).

Related: with this cleanup, `ibc-proto`, `prost`, `prost-types`, and `cosmos-sdk-proto` in `contracts/hydro/Cargo.toml` are only used by `#[cfg(test)]` modules, so they could move to `[dev-dependencies]`.

contracts/hydro/src/contract.rs
dusan-maksimovic: We discussed this before and opted for the whitelist member to be allowed to execute it. In practice, this will be done by the off-chain tool that reads all lockups previously exported from Neutron instance and submits TXs. We agreed it is more flexible if it is not whitelist admin, so we don't have to go through governance proposals. Anyways, if something goes wrong we can discard entire contract easily and start the process again.

p-offtermatt (changes requested):
The one blocker to me is the admin vs whitelisted differentiation, so blocking for now. Otherwise LGTM

contracts/hydro/src/contract.rs
p-offtermatt: Ok good point. In hindsight, I feel like this will just be one very large proposal, and it can go through governance, though maybe it's fine and you are right. We just should make sure to first remove the minting capability before we let users interact with the contract.

contracts/hydro/src/contract.rs
dusan-maksimovic: The process is trusted anyways, so we rely on providing a correct value for LOCK_ID during instantiation. Even if we check `lock_id < LOCK_ID` and some lockups don't get added- then it means we did something wrong. But at that time we could already have the data half-entered, so it is still bad. I would rely on checking the data before submitting the TXs and make sure everything is in order.

p-offtermatt (approved):

contracts/hydro/src/contract.rs
p-offtermatt: sg

contracts/hydro/src/lsm_integration.rs
dusan-maksimovic: Will look into this and clean it up.

dusan-maksimovic: Resolved in https://github.com/informalsystems/hydro/pull/426/commits/e571001aa69c870525784d3da12d089a6b886f2d.
I entirely removed `ibc-proto` from the workspace (it was only used by hydro contract), and `cosmos-sdk-proto` from `hydro` crate, and moved `prost` and `prost-types` into `dev-dependencies` section.

p-offtermatt (approved):

PR #446, opened 2026-08-31
IBC Eureka Inflow adapter implementation

dusan-maksimovic:
## Description

IBC Eureka Inflow adapter smart contract implementation.

---

### Author Checklist

I have...


contracts/inflow/ibc-eureka-adapter-cosmoshub/src/contract.rs
```diff
@@ -0,0 +1,1024 @@
...
+    Ok(Response::new()
+        .add_attribute("action", "toggle_depositor_enabled")
+        .add_attribute("sender", info.sender)
+        .add_attribute("depositor_address", depositor_address.to_string())
+        .add_attribute("enabled", enabled.to_string()))
+}
+
+/// Handle TransferFunds - initiate an IBC Eureka bridge to the destination EVM chain
+#[allow(clippy::too_many_arguments)]
+fn execute_transfer_funds(
```
arlai-mk, line 413: **Fee handling in `TransferFunds`: suggest executor-funded fees, following the CCTP adapter pattern**

Currently `fee_amount` is a free parameter drawn from the adapter's own balance. I'd suggest following the same system as the CCTP adapter instead: the executor attaches the fee via `info.funds` when calling `TransferFunds` (see `cctp-adapter-cosmoshub/src/contract.rs:411-428`), and `fee_amount` is dropped from the message:

```rust
TransferFunds {
    denom: String,
    amount: Uint128,   // net amount the recipient receives
    recipient: String,
}
```

```rust
// Executor sends the Eureka fee in info.funds. It must be in the
// same denom as the bridged token (the Skip entry contract enforces
// this anyway, see handle_eureka_fee in skip-go-cosmwasm-contracts)
let fee_amount = cw_utils::must_pay(&info, &denom)?;
```

Reasons:

1. **Protocol funds stay intact.** Only the bridged net amount leaves the adapter; fees come from the operator side. An oversized fee only wastes the executor's own tokens, since the fee receiver is admin-configured and a compromised executor can't route it to itself.
2. **It closes a drain vector.** With balance-funded fees, a compromised executor EOA can't steal (destinations are allowlisted) but can bleed the adapter's balance by repeatedly bridging with a `fee_amount` of its choosing, since the fee is drawn from depositors' funds each hop. Wallet-funded fees remove that possibility.

Note the balance check stays the same (`amount + fee` against the adapter balance), since `info.funds` are credited to the contract before execution. The CCTP adapter already relies on this. Also suggest keeping `amount` = net received (fee on top), consistent with CCTP semantics. The existing `ZeroFeeAmount` check would be replaced by `must_pay`, which naturally rejects a missing or zero fee payment.

On the executor side this needs no new plumbing: `SkipClient.ts` already has `querySkipEurekaFee`, so a new message builder can mirror `CctpAdapterTransferMessageBuilder` directly (quote, apply buffer, validate against a max fee, attach `funds`).

contracts/inflow/ibc-eureka-adapter-cosmoshub/src/contract.rs
```diff
@@ -0,0 +1,1024 @@
...
+        return Err(ContractError::InsufficientBalance {
+            has: balance.amount,
+            needs: total_transfer_amount,
+        });
+    }
+
+    // Look up destination address from allowlist
+    let destination_address = get_destination_address(&deps.as_ref(), &recipient)?;
+
+    let action_timeout_timestamp = env
```
arlai-mk, line 458: **Timeout handling: document the mixed units, and pass the quote expiry instead of a config-derived fee timeout**

Two suggestions on this block:

1. **Document the units.** The action timeout is in seconds while the fee timeout is in nanoseconds. This matches what Skip:Go emits for this message (the Skip entry contract's `validate_timeout_timestamp` also accepts seconds), so it's correct, but it reads like a bug waiting to be "fixed". Suggest a short code comment and/or a `README` note stating that the action timeout is deliberately in seconds and the Eureka fee timeout in nanoseconds, so a future change doesn't normalize them.

2. **Derive the fee timeout from the quote, not from config.** `eureka_fee_timeout_seconds` is a static admin setting that approximates the expiry of the Skip fee quote. If the configured window is longer than the actual quote TTL (or the executor acts on a stale quote that still fits the window), the fee gets paid but the solver won't match it to a valid quote. The transfer then stalls until the action timeout expires and the recover flow returns the bridged amount, with the fee lost.

Since the executor fetches the Skip quote for the fee anyway, it already has the real `expiration` from `smart_relay_fee_quote`. Suggest adding a per-call `quote_expiry_timestamp` to `TransferFunds` and using it as the Eureka fee timeout, validating on-chain that it's still in the future. The entry contract already rejects calls where `block.time.nanos() > eureka_fee.timeout_timestamp`, so passing the actual quote expiry makes that check meaningful rather than a loose approximation. `eureka_fee_timeout_seconds` can then be dropped from config.

For reference, the EVM side does exactly this: `EurekaHandler` requires `block.timestamp < quoteExpiry` where `quoteExpiry` is the quote's expiration passed per call by the executor.

contracts/inflow/ibc-eureka-adapter-cosmoshub/src/contract.rs
dusan-maksimovic: Addressed in https://github.com/informalsystems/hydro/pull/446/commits/2913303b088b66babf9db35a8528de77573d0129

contracts/inflow/ibc-eureka-adapter-cosmoshub/src/contract.rs
dusan-maksimovic: Addressed in https://github.com/informalsystems/hydro/pull/446/commits/d5d0e3923f6d046d5106568688fa665e75ab8204

arlai-mk:
LGTM!

PR #447, opened 2026-09-01
Add mandatory timeout to SubmitDeployedAmount

arlai-mk:
## Description

Add a mandatory timeout to the `SubmitDeployedAmount` execute message of the Inflow Control Center contract.

`SubmitDeployedAmount` submissions are executed via DAO DAO proposals with one
proposer and one verifier. When the verifier verifies the proposal but its
execution subsequently fails, the proposal remains in the "executable" state
forever. If either the submitter's or verifier's EOA is later compromised, an
attacker could execute a very old, stale proposal and overwrite the deployed
amount with outdated values.

The message now carries a required `timeout` field (a `Timestamp`, i.e. a Unix
timestamp in nanoseconds). Execution reverts with a new
`ContractError::SubmissionExpired` error if `env.block.time` is strictly past
the given timeout. The timeout is checked before any state is written, so an
expired submission cannot affect the deployed amount or fee accrual.

Example message payload (timeout serialized as a base-10 string of nanoseconds):

```json
{
  "submit_deployed_amount": {
    "amount": "123456",
    "timeout": "1759276800000000000"
  }
}
```

**Breaking change:** the message format changed (new required field), so the
control-center contract needs a code upload + migrate before new proposals can
be submitted. Proposals that already passed with the old payload will fail to
execute after the migration (missing required field); a fresh proposal with a
timeout must be submitted instead. This is the intended security property.

---

### Author Checklist

I have...


p-offtermatt (approved):
Great, LGTM!
One extra thing might be to put a "maximum timeout" as a constant, that would avoid the silent failure case where we put way too long of a timeout. But not necessary imo

dusan-maksimovic (approved):
LGTM

PR #448, opened 2026-09-03
Add ability to swap the deposit denom on the Inflow Cosmos Hub vault

arlai-mk:
## Description

Adds a new `SwapDepositDenom` execute message to the Inflow **Cosmos Hub vault**, allowing an
operator to swap the vault's deposit denom (e.g. `USDC.noble` → `USDC.injective`).

The swap is whitelist-gated and is only meaningful for denoms of (basically) the same value,
since vault shares and queued withdrawal amounts are denominated in raw token units and are
not re-priced. The message reverts if the vault contract itself, or **any** registered adapter,
still holds a balance of the current deposit denom, as leftover funds would be stranded once
the denom is swapped. A user deposit landing just before the swap makes the swap fail (the
vault would hold the old denom), so no separate locking of the vault is required. Pending
withdrawal requests store no denom and are simply paid out in the new denom after the swap.

The message is Cosmos Hub only; the Neutron vault returns `Unauthorized`, following the
existing `MintForMigration` precedent.

Most critical files to review:
- `contracts/inflow/vault-cosmoshub/src/contract.rs` (`swap_deposit_denom()` implementation,
  whitelist gate + balance checks and their validation order)
- `contracts/inflow/vault-cosmoshub/src/testing_adapters.rs` (new unit tests)

---

### Author Checklist

I have...


dusan-maksimovic (approved):
LGTM!

p-offtermatt (approved):
LGTM, thanks!

